PR #28, opened 2016-02-29
fix example code

krz:
fixed code to make the example working


jgabry:
Thanks for catching this!

PR #42, opened 2017-03-18
loo expectations with E_loo()

jgabry:
This PR adds a function to the __loo__ package called `E_loo` for computing weighted expectations using loo. It can then be called inside __rstanarm__ but also be used by anyone using the __loo__ package.

To try it out: 

```r
devtools::install_github("stan-dev/loo", ref = "loo_expectation")
library(loo)

# will use rstanarm just to get some stuff to work with here, but rstanarm will 
# provide wrappers to do all this for rstanarm models
library(rstanarm)

# data from help("lm")
ctl <- c(4.17,5.58,5.18,6.11,4.50,4.61,5.17,4.53,5.33,5.14)
trt <- c(4.81,4.17,4.41,3.59,5.87,3.83,6.03,4.89,4.32,4.69)
d <- data.frame(
  weight = c(ctl, trt),
  group = gl(2, 10, 20, labels = c("Ctl","Trt"))
)

fit <- stan_glm(weight ~ group, data = d)
yrep <- posterior_predict(fit)
dim(yrep)

lw <- psislw(-log_lik(fit))$lw_smooth
dim(lw)

E_loo(yrep, lw, type = "mean")
E_loo(yrep, lw, type = "var")
E_loo(yrep, lw, type = "quantile", probs = c(0.1, 0.9))
```

In the __rstanarm__ package the `loo_predict` and `loo_linpred` functions can then call this `loo::E_loo` function internally.


avehtari:
I'll try to check this tonight or tomorrow.

We could add further Pareto-checks, which would take into account the functions. elpd's correspond to zeroth moment, so it was enough to check the moments of the weight distribution. For mean we should check also the moments of w*x, and for variance the we should check the moments of w*x^2. Quantiles are bit more complicated...

jgabry:
@avehtari What do you think of the name `loo_expectation()`? Another possibly that's kind of nice is `E_loo()`.

jgabry:
renamed `loo_expectation()` to `E_loo()`

PR #47, opened 2017-04-27
Add classes waic and psis_loo in addition to loo

jrnold:
Add subclasses `waic`, and `psis_loo` in addition to `loo`

- class of object returned by `waic`  is now `c("waic", "loo")`
- class of object returned by `loo` is now `c("psis_loo", "loo")`
- customized `print` methods for these subclasses

Having these classes makes it easier to work with these object without having to check which elements are in them, but by still having them include the `loo` class, they can still work with `compare`.


jgabry:
Thanks Jeff. I'm on my way out of town for a few days but I will review
these this weekend or early next week.

jgabry:
I totally forgot about this until now (sorry!). But this looks good. Thank you!

PR #61, opened 2018-03-02
Make model weights a generic

jgabry:
This PR makes `model_weights` a generic function with a default method. This will make it easier for packages like rstanarm and brms to provide methods that input a list of models rather than a list of log likelihood matrices. 

@yao-yl The argument name `log_lik_list` has been changed to `x` to make this less confusing if rstanarm and brms use lists of models. 

@paul-buerkner you should be able to define a model_weights method now


jgabry:
@paul-buerkner Actually, now that I think about it, it might make more sense for brms and rstanarm to provide methods functions for creating the log-likelihood lists for passing to loo::model_weights, rather than implementing model_weights methods. Since model_weights wants a list, it doesn't make sense to have a model_weights.stanreg or model_weights.brmsfit since those would expect the first argument to be a fitted model object rather than a list of them. What do you think?

paul-buerkner:
I would prefer having a method that calls model_weights directly rather than a method which produces a list of log-likelihoods to be passed to model_weights afterwards. Please see the new-loo branch of brms (brmsfit-methods.R and loo-helpers.R) for the way I currently build a wrapper around model_weights. Also I think that the name model_weights and model_select is too general for use as a method in rstanarm or brms. I prefer loo_weights and loo_select as the method names, but that's totally up to you of course.

jgabry:
Ok thanks for the feedback. Let me think about this and get back to you.

paul-buerkner:
Or one might stick to the name 'model_weights' in which case, I may add so more functionality to support other weight types, as well within the method.

PR #63, opened 2018-03-22
model_weights --> loo_model_weights

jgabry:
This PR does the following:

- renames `model_weights` to `loo_model_weights` (both the generic and default method) 
- removes the `seed` argument (it already respects `set.seed`) 

@paul-buerkner I think the only change you need to make to your PR in brms is to replace `model_weights` with `loo_model_weights`


paul-buerkner:
That's great! Thank you! One small question: What happend to the `model_select` function. It seems to have been removed from the branch?

paul-buerkner:
That's great! Thank you! One small question: What happend to the `model_select` function. It seems to have been removed from the branch?

jgabry:
I think that one was never really intended to be released but I forgot to
remove it when reviewing Yuling’s branch. Then I removed it at some point
in the past few weeks without realizing you had already defined a method
for it. Sorry about that. Anyway, for now we’re just going to have the
loo_model_weights function.

paul-buerkner:
No worries. Just wanted to make sure I am not missing something.

PR #77, opened 2018-04-23
Vignette about loo for non factorizable models

### LINKED ISSUE #75: Vignette: LOO for non-factorizable joint normal models
Just opening an issue to have a record of this on the to-do list. @avehtari and @paul-buerkner are already working on it. 

(This is related to the issue of single data points raised in #71.)

jgabry:
closes #75 

Unfortunately this substantially increases the amount of time it takes to install or check the package if not ignoring the vignettes. But I think it's worth having.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ loo.Rproj
 .Rproj.user/
 .Rproj.user
 # produced vignettes
+vignettes/loo2-non-factorizable_cache/*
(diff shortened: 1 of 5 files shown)


jgabry:
Since this was already reviewed by all of us I'll just merge this. The website will also need to be updated separately. Right now this is just to merge the code into the master branch.

PR #90, opened 2018-10-11
add LFO vignette

paul-buerkner:
This PR adds a vignette about leave-future-out cross-validation based on our case study.

### DIFF
--- a/vignettes/loo2-lfo.Rmd
+++ b/vignettes/loo2-lfo.Rmd
@@ -26,20 +26,6 @@ knitr::opts_chunk$set(
 )
 ```
 
-
-```{r pkgs, cache=FALSE}
-library(brms)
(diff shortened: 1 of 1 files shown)


avehtari:
Is this the same as in html in m-step repo? No need to knit this version?

paul-buerkner:
I also have the knitted version ready on a branch, but it seems as if the loo master branch contains no knitted version of any vignette. That's why I didn't include it. Would love to have @jgabry in this discussion to tell us about where/when to add the knitted version.

paul-buerkner:
After some discussions with Aki and some further changes, this vignette is ready to be included.
I will later move it to gh-pages in order to update the website accordingly.

jgabry:
Thanks Paul!

PR #92, opened 2018-12-18
psis-loo using posterior approximations

MansMeg:
Hi!

Here is a PR containing a new function to estimate k_hat based on posterior approximations such as ADVI and Laplace. There are four files that are new/has changes.

1. DESCRIPTION
I added the checkmate package. It is a package for argument assertions. I think it is probably the best R package for function arguments assertions with very clear assertions. You can see the usage in the main R function.

2. R/psis_approximate_posterior.R
This file contains the function used and it returns either a PSIS object or a LOO object depending on whether log_liks are supplied or not. I hope the api is clear and contains everything that it should contain.

3. data-raw/generate-example_posterior_approximation.R
This file contain information how the test data is generated. It stores the data in the testthat folder.

4. tests/testthat/test_data_psis_approximate_posterior.rda
This file contain testdata. It is currently 500 kb large, I hope that is ok, but the file will not change (hopefully).

5. tests/testthat/test_psis_approximate_posterior.R
A test suite that test the function for ADVI and Laplace for correlated and uncorrelated posterior as well as skewed and fat tailed posteriors.

Just let me know if I misunderstood anything.

With kind regards
Måns


jgabry:
Thanks @MansMeg! I'll take a look as soon as possible.

R/psis_approximate_posterior.R
```diff
@@ -0,0 +1,46 @@
...
+#' @return
+#' If log likelihoods are supplied, the function returns a \code{loo} object,
+#' otherwise the function returns a \code{psis} object.
+#'
+#' @seealso \code{loo} and \code{psis}
+#'
+#' @template loo-and-psis-references
+#'
+#' @export
+psis_approximate_posterior <- function(log_p, log_q, log_liks = NULL, cores, save_psis, ...){
```
jgabry: I think there should be explicit defaults for the `cores` and `save_psis` arguments, e.g.,

```suggestion
psis_approximate_posterior <- function(log_p, log_q, log_liks = NULL, cores = getOption("mc.cores", 1), save_psis = FALSE, ...){
```

R/psis_approximate_posterior.R
```diff
@@ -0,0 +1,46 @@
+#' Diagnostics for Laplace and ADVI approximations and Laplace-loo and ADVI-loo
+#'
+#' @param log_p The log-posterior (target) evaluated at samples from the proposal distribution (q).
+#' @param log_q The log-density (proposal) evaluated at samples from the proposal distribution (q).
```
jgabry: Can you clarify in the doc that `log_p` and `log_q` should be vectors (right?) and  the same length?

R/psis_approximate_posterior.R
```diff
@@ -0,0 +1,46 @@
...
+    approx_correction <- approx_correction - max(approx_correction) # Handle underflow/overflow
+    log_ratios <- matrix(approx_correction, ncol = 1)
+  } else {
+    log_ratios <- -log_liks
+    for(j in 1:ncol(log_ratios)){
+      log_ratios[,j] <- log_ratios[,j] + approx_correction
+      log_ratios[,j] <- log_ratios[,j] - max(log_ratios[,j]) # Handle underflow/overflow
+    }
+  }
+  psis_out <- psis.matrix(log_ratios = log_ratios, cores = cores, r_eff = rep(1, ncol(log_ratios)))
```
jgabry: I think you can replace `r_eff = rep(1, ncol(log_ratios))` with `r_eff = NA`. I think I enabled `NA` as a way to have it ignore the `r_eff` argument without throwing a warning. But no big deal either way here. 

```suggestion
  psis_out <- psis.matrix(log_ratios = log_ratios, cores = cores, r_eff = NA)
```

R/psis_approximate_posterior.R
```diff
@@ -0,0 +1,46 @@
...
+
+  if(is.null(log_liks)){
+    approx_correction <- approx_correction - max(approx_correction) # Handle underflow/overflow
+    log_ratios <- matrix(approx_correction, ncol = 1)
+  } else {
+    log_ratios <- -log_liks
+    for(j in 1:ncol(log_ratios)){
+      log_ratios[,j] <- log_ratios[,j] + approx_correction
+      log_ratios[,j] <- log_ratios[,j] - max(log_ratios[,j]) # Handle underflow/overflow
+    }
```
jgabry: If there are a lot of data points then maybe it makes sense to allow parallelization of this loop? 

We could also replace the loop with matrix operations, e.g., 
```r
log_ratios <- log_ratios + approx_correction
log_ratio_max <- apply(lr, 2, max)
log_ratios <- sweep(log_ratios, MARGIN = 2, STATS = log_ratio_max)
```
 but maybe that requires too much memory for many observations?

R/psis_approximate_posterior.R
MansMeg: Sure! I assumed since the function will not be a part of the public API I assume you did not want any defaults. In my experience, non-API functions using defaults can miss usage bugs in accessing the function, and hence not using defaults will be more defensive.  But you know how you want it so I do what you think is best and is in line with the overall structure of the project. Just let me know how you want it.

R/psis_approximate_posterior.R
MansMeg: Yes! Ill do that!

R/psis_approximate_posterior.R
MansMeg: Aki made the same comment and I tried it but it actually did not work. Maybe I'm in some old branch. I can do it, but then the current test suite will fail.

Just let me know how you want it.

R/psis_approximate_posterior.R
MansMeg: I'll fix this. I totally forgot the sweep function here. Good suggestion!

R/psis_approximate_posterior.R
jgabry: Oh, I was thinking this would maybe be exposed to users, but if it won't be part of the public API then I agree. It's fine to leave it without defaults.

R/psis_approximate_posterior.R
jgabry: It's fine as is, don't worry about it.

R/psis_approximate_posterior.R
MansMeg: Yes. Already fixed.

MansMeg: If I remember correctly we said that I should implement it as a standalone function and then you could plug it in with the laplace objects since there are currently just a list.

PR #93, opened 2018-12-20
compare --> loo_compare

jgabry:
This PR deprecates the `compare()` function in favor of `loo_compare()`. This is for several reasons: 

* `compare()` is not a generic with methods, which results in rstanarm, brms, etc., using different names for wrappers around `compare()`. With `loo_compare()` we can have methods `loo_compare.stanreg`, `loo_compare.brmsfit`, etc, unifying the names. 

* `compare()` also conflicts with function names in other packages (e.g. testthat::compare)

We should time the release of this so that the `loo_compare()` methods in brms and rstanarm are released at the same time.


paul-buerkner:
One little thing: Could we add an `is.kfold` function as well?

jgabry:
I'm ready to merge this but @avehtari do you want to first try out the `loo_compare()` function (on the `loo_compare`) branch and let us know if you like the interface/output?

jgabry:
@paul-buerkner I removed the deprecation warning from `compare()` for now as we discussed. I'm running into a few issues with reverse dependency checks and I need to sort out if they are loo related, but otherwise I think this is ready to go.

PR #95, opened 2018-12-30
use code from RStan monitor.R to compute MCMC ESS, fixes #85

### LINKED ISSUE #85: relative_eff uses slightly wrong algorithm
`relative_eff` computation in R has two problems: 
1. initial sequence truncation is based on individual lags, but should be based on pairs as in Stan
2. autocorrelations are used differently, which may be suboptimal compared to FFT version used in Stan

Fix
 - [x] use Geyer's initial monotone sequence as in Stan
 - [x] check that autocorrelation values match with code in Stan or change the code to use FFT

avehtari:
Fixes #85 

Code copied from RStan monitor.R to have a consistent behavior in loo
- [x] autocorrelation computations uses FFT
- [x] use Geyer's initial monotone sequence

Code in monitor.R has been previously tested to produce expected results.
Only internal functions have been added and removed.
This is also 3 times faster due to use of FFT (from 15s -> 5s for 4000 by 262 log-likelihood matrix)

Copyright as in RStan monitor.R.
Llicense in RStan monitor.R is compatible with loo package.


jgabry:
I just pushed a fix for some tests but this looks good otherwise.

Issue #49, opened 2017-06-07
k-fold-cv helper functions


bnicenboim:
Adapt helper functions [from here](http://htmlpreview.github.io/?https://github.com/stan-dev/stancon_talks/blob/master/2017/Contributed-Talks/07_nicenboim/kfold.html) to deal with k-fold CV.  (See also #25).

The user should add to the data declaration a vector that indicates which observations go to training (`trainidx`).

And the models should look like this:
```
model {
...
for (n in trainidx) 
      target += model_lpdf[n];
}
```

and some models could be written simply as

```
model {
...
y[trainidx] ~ normal(mu, sigma);
}
```

And the generative quantities has a `vector[N_obs] log_lik`. My helper functions extract the relevant log_lik, and calculate the elpd like it is done with loo.

avehtari:
Thanks! Your case study looks great, and this kind of approach would be useful to add to the package.

jgabry:
Following up on this finally! To start with we'll have the following options for a helper function that prepares data subsets for K-fold CV:

* [ ] Random permutation (ok when exchangeable, i.e. when LOO ok but N is big or k-hat is big)
* [ ] Balanced division based on outcome variable (e.g. if we have binary outcome with not so many obs in one of the categories then a random permutation can result in separation)
* [ ] Stratification based on discrete covariate(s)

avehtari:
I comment here although implementation goes to rstanarm. `kfold` in rstanarm should accept the index array from these functions.

jgabry:
Yeah, this is related to https://github.com/stan-dev/rstanarm/issues/228

avehtari:
It seems there is a need also for kfold splits for time series. I don't know what what would a good shorter name for `kfold_split_m_step_ahead-prediction`? Maybe `kfold_split_timeseries', with options for how many steps ahead and a gap steps (which could be infinite for not including anything after the prediction steps).

jgabry:
For the observations that are not in the testing set and also not in the training set (which depends on tuning parameters), should we go with `NA`?

avehtari:
- [ ] It seems our kfold helper function names don't match with https://scikit-learn.org/stable/modules/cross_validation.html and it's likely to be my mistake. We should fix these to follow what they have (they have also nice figs)
- [ ] the current balanced allows only binary x, but it would be useful to allow any categorical x

ping @jgabry

PR #96, opened 2019-01-03
fix k-fold-cv helper functions

avehtari:
As mentioned in https://github.com/stan-dev/loo/issues/49#issuecomment-449086744

- [x] Make our kfold helper functions to match with https://scikit-learn.org/stable/modules/cross_validation.html
- [x] Allow non-binary discrete x for `kfold_split_stratified`

### DIFF
--- a/R/kfold-helpers.R
+++ b/R/kfold-helpers.R
@@ -7,9 +7,8 @@
 #' @param K The number of folds to use.
 #' @param N The number of observations in the data.
 #' @param x A discrete variable of length \code{N}. Will be coerced to
-#'   \code{\link{factor}}. For \code{kfold_split_balanced} \code{x} should be a
-#'   binary variable. For \code{kfold_split_stratified} \code{x} should be a
-#'   grouping variable with at least \code{K} levels.
+#'   \code{\link{factor}}. For \code{x} should be a grouping variable
+#'   with at least \code{K} levels.
 #' @return An integer vector of length \code{N} where each element is an index
 #'   in \code{1:K}.
 #'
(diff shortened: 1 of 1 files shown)


R/kfold-helpers.R
```diff
@@ -75,27 +71,28 @@ kfold_split_random <- function(K = 10, N = NULL) {
...
-    length(unique(x)) == 2,
     K == as.integer(K),
     length(K) == 1,
     K > 1,
     K <= length(x)
   )
-  x <- as.integer(as.factor(x)) - 1
-  x0_id <- which(x == 0)
-  x1_id <- which(x == 1)
-  bins <- rep(NA, length(x))
```
jgabry: I think you still need  `bins <- rep(NA, length(x))` otherwise later you can't do `bins[xids] <- rep(1:K, ceiling(N/K), )[1:N]` because `bins` doesn't exist.

R/kfold-helpers.R
```diff
@@ -75,27 +71,28 @@ kfold_split_random <- function(K = 10, N = NULL) {
...
-  bins[x0_id] <- kfold_split_random(K = K, N = length(x0_id))
-  bins[x1_id] <- kfold_split_random(K = K, N = length(x1_id))
+  x <- as.integer(as.factor(x))
+  Nlev <- length(unique(x))
+  N <- length(x)
+  xids <- numeric()
+  for (l in 1:Nlev) {
+      xids <- c(xids, which(x==l)[sample.int(sum(x==l))])
```
jgabry: I think 
```r 
which(x==l)[sample.int(sum(x==l))]
```
is equivalent to 
```r 
sample(which(x==l), size = 1)
```
right? If so then I think the latter makes the code easier to understand.

R/kfold-helpers.R
```diff
@@ -75,27 +71,28 @@ kfold_split_random <- function(K = 10, N = NULL) {
...
+  }
+  bins[xids] <- rep(1:K, ceiling(N/K), )[1:N]
```
jgabry: This is where there there's error I mentioned above because`bins` doesn't exist.

R/kfold-helpers.R
avehtari: without size=1 is correct

avehtari:
Sorry for making a pull request with broken code. I modified tests, too, as I need to learn to use tests.

Issue #81, opened 2018-04-27
Add more doc about interpreting loo output: approx SE of elpd_diff, p_loo, k


jgabry:
From Discourse: 

http://discourse.mc-stan.org/t/loo-comparison-in-reference-to-standard-error/4009/4?u=jonah

avehtari:
Glossary merged

PR #98, opened 2019-01-31
Add glossary page to package documentation

jgabry:
@avehtari I've added a page to the doc with your definitions (minor edits). It will be available from R using `help("loo-glossary")` and once released it will be available at 
https://mc-stan.org/loo/reference/loo-glossary (not live yet) so we could link to it from anywhere. 

You can ignore all the files in the PR except `loo-glossary.R`. The others are just other doc pages that got regenerated. 

fixes issue #81


avehtari:
Looks good. I made some additional doc edits also to some other files. If they are ok for you, then you can merge.

Issue #94, opened 2018-12-30
Windows 10 parallelization can be super slow


avehtari:
Lawrence Hunsicker reported that Windows 10 is super slow in parallelization if `.rprofile` has `options(mc.cores = parallel::detectCores())`. The same line in a script is not a problem,
https://discourse.mc-stan.org/t/problem-parallelizing-rstanarm-loo/6433/10

I don't know what we can do for this in `loo` package, but I add this here if someone is hit by the same Windows bug.

jgabry:
We could mention this in the package documentation, but if this is very severe then maybe it's worth printing a message about it to Windows users when the package is loaded?
I'm not sure what the message would be precisely, but it's possible by adding to the `.onAttach()` function. Currently it's this:

https://github.com/stan-dev/loo/blob/e147fd32987bbc619c96537c8f8e043febb4125d/R/zzz.R#L1-L10

PR #99, opened 2019-02-03
add warning about Windows 10 core problem #94

avehtari:
add package load message warning about Windows 10 core problem #94


jgabry:
Thanks @avehtari. I just changed it so that it won't print the warning unless it detects a Windows operating system. No need to print it if the user isn't on Windows.

PR #103, opened 2019-03-18
more warnings for loo_compare

### LINKED ISSUE #102: More warnings for loo_compare
_Originally posted by @avehtari in https://github.com/stan-dev/rstanarm/pull/340#issuecomment-473955094_:

After additional thinking and testing, it would useful to allow but give a
warning for `loo_compare(kfold_1, loo_2)` and `loo_compare(kfold_1, kfold_2)` if K is not the same and for loo K=n. Warning like "Warning: comparing 10-fold-CV to loo. For more accurate comparison use the same number of folds or loo for all models compared."  (in the second case, e.g., ...comparing 10-fold-CV to 20-fold-CV.)

jgabry:
closes #102

### DIFF
--- a/R/loo_compare.R
+++ b/R/loo_compare.R
@@ -147,6 +147,7 @@ se_elpd_diff <- function(diffs) {
 #' @param loos List of loo objects
 #' @return Nothing, just possibly throws errors/warnings
 loo_compare_checks <- function(loos) {
+  ## errors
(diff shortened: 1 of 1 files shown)


jgabry:
@paul-buerkner You don't really need to review this, just wanted you to be aware of these new warning messages from `loo_compare()`

PR #110, opened 2019-06-03
Subsetting psis objects

jgabry:
The PR introduces a `subset()` method for `"psis"` objects. This is useful if you want to drop some of the contents of a `"psis"` object without breaking it. For example, **bayesplot** has an option for this for `ppc_loo_intervals()` (https://mc-stan.org/bayesplot/reference/PPC-loo.html?q=subset#arguments) implemented via an internal function but it would be good if **loo** provided this functionality and **bayesplot** can call it.


avehtari:
@jgabry is this still work in progress?

florence-bockting:
We may address this within the refactoring PR #363, in which case we would move the discussion there and close this PR.

PR #113, opened 2019-07-05
Add loo_subsample and loo_approximate_posterior

MansMeg:
Hi!

Here come loo_subsample and loo_approximate_posterior functionality.

Implementation question/issues
1. I have not implemented everything we may want to need now. What is currently missing is approximations with truncated IS and subsample_loo for loglik arrays and matrices. Those should be easy to implement, although the current version has the basics.
2. I have now implemented ndraws() and nparameters() functions for loo_subsample(). Those may be of more general use and can the be moved somewhere else. Now they are in loo_subsample.R.
3. I have not profiled the code so it may be possible to improve performance.
4. I now put most in the loo_subsample.R and the loo_approximate_posterior.R files.
5. I have not bumped the DESCRIPTION file 

Technical questions/issues
1. I have now removed r_eff from loo_approximate_posterior(). I could not think of a situation with approximate posteriors and autocorrelation?
2. Now r_eff is implemented as in ordinary loo, but it makes sense to only use r_eff for the subsampled observations. But I could not find a good way of doing this.
3. I'm currently not handling MCSE correctly. Should I remove it for subsample_loo() return it only for the subsampled observations or for the totals? I'm not 100% sure what to do there.
4. I would love to have a code review of how I implemented the naive diff SE estimation (for different subsamples) at row 77-96 in loo_compare.psis_loo_ss_list.R and how I compute looic at row 1083-1085 in loo_subsample.R . Maybe @avehtari could look at this?


paul-buerkner:
Hi Mans, 

I have looked at the ssloo PR and have a couple of comments/questions.

(1) The current internal structure of `loo_subsample.function` is likely not going to work with brms. The reason is that you assume certain operations to be possible on the `draws` object passed by the user/other packages that cannot be canonically implemented for more complex objects (such as `brmsdraws` objects) without huge additional efforts. 

- `thin_draws` is easy to implement for matrices but not for nested list-like objects whose posterior draws happen to be stored in a lot of different places. Why do we need to thin the draws at all in the process? Wouldn't it be sufficient if the user, given that they wanted so, just passed a `draws` object which was thinned already before passing it to `loo_subsample`?

- `compute_point_estimate` assumes that a point estimate can be obtained from the draws object, but this is not reasonably possible (without huge code overhead) for complex list-like draws objects for the above reason. 

- Why do we need `nparameters` if we do subsampling loo on MCMC draws and do not compute any approximate posteriors? Which parameters are to be counted and which are not?

- I have got it to work with brms when using `loo_approximation = "waic"` and dummy functions for `ndraws`, `thin_draws` and `nparameters` but this seems very workaroundy to me.

(2) Contrary to what I said a few weeks ago, I think that adding all the approximate posterior arguments to loo_subsample is actually quite confusing. Users you simply work with MCMC samples will not really know which arguments are required in `loo_subsample` and which can safely be ignored. We either need to improve the documentation in that regard or split up the functions so that `loo_subsample` only deals with MCMC samples and `loo_approximate_posterior` with actual posterior approximations requiring log_p, log_q, and the like (I prefer splitting them up if this is not too complicated). If we decide to keep these arguments in `loo_subsample` they should be moved towards the end and do not come before much more relevant arguments such as `loo_approximation`, `estimator` etc.

(3) The documentation requires more explanation of why we actually need `loo_subsample` and what (on a conceptual basis) actually happens behind the scenes. Also, an example should be added to guide users with the recommended workflow. Lastly, the order of the arguments in the doc seems off at times, but this may very well be a local problem for me when building the package.

(4) What is the reason behind using plpd as the default of  `loo_approximation`? 

(5) With the generics of `ndraws` etc being kept internal, I need to explicitely export the corresponding methods for brms-related objects, as I cannot register them on an internal generic of another package (to my knowledge). Thus, we need to export these generics, but if we do it could make more sense to put them into the **rstantools** package rather than in **loo** as they may not just be of relevance for loo related functionality.

(6) I also don't see r_eff being relevant for an approximate posterior. 

(7) I think the problem of computing r_eff only for the relevant observations comes down to a more general problem of handling r_eff in pointwise evaluation. Here, we do not have the full log_lik matrix beforehand to compute r_eff, but instead compute the pointwise log_lik internally on the fly. Thus, r_eff needs to be computed internally on the fly as well. @jgabry What was your reason of requring users to compute r_eff themselves instead of having it computed internally in loo? The latter seems easily possible as we have the log_lik values available internally regardless of whether we do use the .function or .matrix approach.

Ok that's it from me for now. Hope this helps in moving these features forward.

MansMeg:
This is great @paul-buerkner! 

I'll try to answer these questions as much as possible,

> (1) The current internal structure of loo_subsample.function is likely not going to work with brms [...] without huge additional efforts.

Yes. I know this may be a problem and I tried to work around it the best I could, so let’s see how we can work this through. loo_subsample() has two main parts
a. Approximate loo_i:s as good as possible for all observations
b. Compute (exact) psis loo on a subsample (and then estimate the total). 
The better the approximation, the fewer samples are needed.

> thin_draws is easy to implement for matrices but not for nested list-like objects whose posterior draws happen to be stored in a lot of different places. Why do we need to thin the draws at all in the process? Wouldn't it be sufficient if the user, given that they wanted so, just passed a draws object which was thinned already before passing it to loo_subsample?

This is a good question. So using loo_approximate_draws and using methods that approximate loo using the whole posterior, such as lpd, waic and truncated IS, thinning will reduce the computational burden of the approximation (step a) above). Although, we still want to use the full posterior for the psis loo subsample. So supplying different draws objects could work but would need that we compute the loo approximation and the subsample separately, and this is not implemented right now.

I suggest that we, right now, don't support loo_approximate_draws for brms. Partly because loo_approximate_draws is more research-oriented anyway, and the results are not very good. Then this function will not be needed, it should just return the draws object without doing anything, except throwing a warning. 

> compute_point_estimate assumes that a point estimate can be obtained from the draws object, but this is not reasonably possible (without huge code overhead) for complex list-like draws objects for the above reason.

Ok, this may be more important. The point estimates are used for the default approximations for maximum performance so I think this need to be implemented. I guess you have some kind of parameter mean computation somewhere? Then just slabbing the mean on the the draws object as an attribute and then extract it in the point estimate function should suffice. Or am I missing something here?

> Why do we need nparameters if we do subsampling loo on MCMC draws and do not compute any approximate posteriors? Which parameters are to be counted and which are not?

This is only used to assert that the draws object is the same when update.loo_subsample() is used. So it could return any definition of parameter - preferable the easiest one.

> I have got it to work with brms when using loo_approximation = "waic" and dummy functions for ndraws, thin_draws and nparameters but this seems very workaroundy to me.

I do not think this is particularly bad habit to do what you have done - thats actually my whole intention. =) It is clear what the loo_subsample() needs from brms, or any higher-layer package, and as long as brms conforms and supply those methods, everything should work out-of-the box. If you workaround and don't implement something, you can hinder that to be a part of the brms API or throw a warning. Although, I may definitely be wrong here and you might have a better solution? Maybe I should document exactly what these generics are used for, to make it more clear what the purpose is for higher level functionality.

> (2) Contrary to what I said a few weeks ago, I think that adding all the approximate posterior arguments to loo_subsample is actually quite confusing. Users you simply work with MCMC samples will not really know which arguments are required in loo_subsample and which can safely be ignored. We either need to improve the documentation in that regard or split up the functions so that loo_subsample only deals with MCMC samples and loo_approximate_posterior with actual posterior approximations requiring log_p, log_q, and the like (I prefer splitting them up if this is not too complicated). If we decide to keep these arguments in loo_subsample they should be moved towards the end and do not come before much more relevant arguments such as loo_approximation, estimator etc.

I think combining posterior approximations and subsampling is good, and something users would like to be able to do. Maybe improve documentation and move the functions toward the end would be the best solution? Another would be to just have an ellipsis, and if log_p and log_g is supplied it uses that? That would though be a less clear user interface.

> (3) The documentation requires more explanation of why we actually need loo_subsample and what (on a conceptual basis) actually happens behind the scenes. Also, an example should be added to guide users with the recommended workflow. Lastly, the order of the arguments in the doc seems off at times, but this may very well be a local problem for me when building the package.

I can definitely fix this. I have written a vignette, should it be included there, in the R doc or in both?

> (4) What is the reason behind using plpd as the default of loo_approximation?

It is the less computationally costly approach and works really well in our experiments. It also has the best scaling properties of all non-gradient methods.

> (5) With the generics of ndraws etc being kept internal, I need to explicitely export the corresponding methods for brms-related objects, as I cannot register them on an internal generic of another package (to my knowledge). Thus, we need to export these generics, but if we do it could make more sense to put them into the rstantools package rather than in loo as they may not just be of relevance for loo related functionality.

Yes. Is it a problem to export a method? I agree that ndraws etc may be a good generic for many packages. So maybe put it in an rstantools package? I have. Not thought much about this. I just realized that some generics must be very common.

> (6) I also don't see r_eff being relevant for an approximate posterior.

Great!

> (7) I think the problem of computing r_eff only for the relevant observations comes down to a more general problem of handling r_eff in pointwise evaluation. Here, we do not have the full log_lik matrix beforehand to compute r_eff, but instead compute the pointwise log_lik internally on the fly. Thus, r_eff needs to be computed internally on the fly as well. @jgabry What was your reason of requring users to compute r_eff themselves instead of having it computed internally in loo? The latter seems easily possible as we have the log_lik values available internally regardless of whether we do use the .function or .matrix approach.

Yes. This is also something I wonder.

> Ok that's it from me for now. Hope this helps in moving these features forward

It does indeed! Thank you Paul!

paul-buerkner:
Here are my notes from a call with @MansMeg discussing the points raised above:

- `thin_draws` can be a method that does not actually thin (if this is impossible or hard for the provided object) but should throw a warning if an attempt is made to do it anyway.
- `compute_point_estimate` has the right API in the sense that it gives the burden back to the developer of the packages interfacing loo to return an object just containing the point estimates (in whatever format the developer decides) as long as it is compatible computing the provided log_lik computing function.
- `nparameters` should be replace by something else to better capture the identity of the model so that it can be used as a proper check in `update.loo_subsample` if the provided model is the same as the original one.
- The above mentioned generics should be exported by **loo** so that the related methods (for brmsfit objects, say) can be properly registered as S3 methods on these generics. Perhaps we should add the doc keyword `@internal` to the generics so that they don't appear in the doc index and should change their names to `.<generic name>` to indicate (via the leading .) that they are usually not ment to be called by users.
- Keeping the details in the vignette is find as long as its properly pointed to in the doc.

paul-buerkner:
I think we are getting very close to having this PR ready. With the current `loo_subsample` method of brms I see for some models the following error: 

```
Error: 'observations' is larger than total sample size.
```

I cannot tell whether this happens because of brms or loo but I think it would be good to figure this out before merging. Here is a reproducible example:

```R
devtools::install_github("paul-buerkner/brms", ref = "ssloo")
library(brms)
fit1 <- brm(count ~ zAge + zBase * Trt + (1|patient),
             data = epilepsy, family = poisson())
loo_subsample(fit1)
```

MansMeg:
Great. I just pushed some final fixes for the documentation, generics etc. Hopefully this will also fix the two failing test cases, but that remains to be seen.

The problem you encountered is that the default subsample size is 400 and the epilepsy dataset contain roughly 250 observations. Now it throws an error. I tried to make the error more clear. But maybe we should just set the observation to the full dataset and trow a warning instead?

paul-buerkner:
Ah, this makes sense. Sorry this was my bad. I think the error is clear. Perhaps one could say **Argument*' 'observations' to make clear this is something the user can simply change?

paul-buerkner:
Ok, the PR looks good to me now and is ready to be merged.

@jgabry, @avehtari any objections to merging this?

paul-buerkner:
After @avehtari gave his ok yesterday, we are ready to merge this PR.

Thank you @MansMeg for your work! I think this is a really nice addition to the loo package!

jgabry:
I forgot to answer this good question from @MansMeg:

> @jgabry What was your reason of requring users to compute r_eff themselves instead of having it computed internally in loo? The latter seems easily possible as we have the log_lik values available internally regardless of whether we do use the .function or .matrix approach.

It’s only possible if we know the chain_id for each draw, which we do in the array case (possibly the function case), but not the matrix case without the user supplying a chain_id variable. So I think @avehtari and I decided that loo methods in other packages like rstanarm and brms can compute r_eff but the primary loo methods in loo would take r_eff as an input. 

We could change this (I think without breaking backwards compatibility) if there’s a good reason.

MansMeg:
Ah. Of course, it totally makes sense. I assumed we "knew" the chain structure.

Issue #112, opened 2019-06-12
Incomplete documentation for psis function


dpsimpson:
This came up when I was trying to use the `weights.psis` function to actually make an importance sampler. The weights that are returned are _not_ the log-PSIS weights, but rather the log-PSIS weights _minus_ the largest log ratio.  @avehtari documented the solution here. https://discourse.mc-stan.org/t/trouble-using-loo-psis-for-importance-sampling/9209/2

This is not super urgent, but it would be convenient to have it clarified in a future version.  Also possibly to have a simple vignette that shows how to implement an importance sampler.  I'm happy to volunteer to do these things (but not in the next week or so).

avehtari:
To further clarify the log-PSIS weights (non-translated ones) are useful if
1) if normalization terms are known and regular importance sampling (non-self-normalized) is used
2) if normalization terms are unknown and need to estimate the ratio of normalization terms (e.g. one of the normalization terms is known, and then we want to infer the other one)

In PSIS-LOO and in most other self-normalized importance sampling, we don't need these and it was easier to leave the non-normalized weights as they are now. I now think that we could add back that largest log ratio, although then we need to check that we are not exponentiating those log-PSIS weights before subtracting the max log weight.

jgabry:
Closing since #146 was merged

PR #123, opened 2019-10-06
Add TIS and IS as importance sampling methods

MansMeg:
Hi!

I'm currently running experiments with TIS and IS and @avehtari suggested that I should include it in the loo package. I struggled a little to how to implement them but this is the minima-maintenance version. The problem is that the psis() function now can return objects that used tis or is. The alternative would be to implement tis() and is() that would be a lot of copied code. There are tests included. I suggest that @avehtari double-check the `do_tis_i()` and `do_is_i()`.


avehtari:
To clarify, TIS doesn't need to be exposed, but it could be used by subsampling approaches.

MansMeg:
Indeed. Now it is an argument in `psis()` with default "PSIS". So it is exposed, but not from `loo()`. Is that what you mean?

R/psis.R
```diff
@@ -224,26 +234,34 @@ psis_object <-
...
   S <- nrow(log_ratios)
   tail_len <- n_pareto(r_eff, S)
-  throw_tail_length_warnings(tail_len)
+
+  if(is_method == "PSIS") {
+    is_fun <- do_psis_i
+    throw_tail_length_warnings(tail_len)
+  }
+  if(is_method == "TIS") is_fun <- do_tis_i
+  if(is_method == "IS") is_fun <- do_is_i
```
avehtari: maybe use if else if -structure to make it clear that only one of these can happen, and error if is_method is undefined?

R/psis.R
```diff
@@ -330,6 +349,48 @@ do_psis_i <- function(log_ratios_i, tail_len_i) {
...
+#'   log likelihoods).
+#' @param tail_len_i Not used. Included to conform to PSIS API.
+#'
+#' @details Implementation of Truncated importance sampling (PSIS), a method for
+#' stabilizing importance ratios. The version of TIS implemented here
+#' corresponds to the algorithm presented in Ionides (2008) with truncation at
+#' sqrt(S).
+#'
+#' @return A named list containing:
+#' * `lw`: vector of unnormalized log weights
```
avehtari: it might be useful to state that unnormalized and sifted to have max=0, or change the function to not to do sifting.

avehtari: see #112

avehtari:
Maybe we should now fix also #112 with this PR (except the vignette)

R/psis.R
MansMeg: Indeed! Ill fix.

R/psis.R
MansMeg: Indeed. I did it for it to be similar to the PSIS implementation. I update the doc and fix issue #112.

MansMeg:
Now the three suggestions should have been fixed. I think I also fixed #112, double-check if it is sufficient.

avehtari:
#112 says
>  I now think that we could add back that largest log ratio, although then we need to check that we are not exponentiating those log-PSIS weights before subtracting the max log weight.

Your fix did not do that.

avehtari:
Or can weights method return also TIS and IS weights? And that would be recommended way to call those?

MansMeg:
So to fix #112 I just added that the returned weight sare the weights minus the largest log ratio - to make it explicit in the documentation.

Yes. So now a psis object is returned with TIS, PSIS and IS and then we can extract the weights with weights().

avehtari:
> Yes. So now a psis object is returned with TIS, PSIS and IS and then we can extract the weights with weights().

I guess it's optionally with one of these and not all all the time?

MansMeg:
Actually the smoothing and truncation (and removing max value) are done when creating the psis object. The weights-function only extract and exponentiate or normalize the weights. So you choose to use TIS when calling `psis()` then the weights in the psis object will be truncated (or Pareto-smoothed if you do not choose anything).

avehtari:
Your explanation was not clear enough, and I want to completely certain and check that also documentation is clear on this.

> Actually the smoothing and truncation (and removing max value) are done when creating the psis object.

This sentence can be interpreted so that all PSIS, TIS and IS are done when creating psis object.

> So you choose to use TIS when calling psis()

Why not call tis(), or does this do both TIS and PSIS?

> then the weights in the psis object will be truncated (or Pareto-smoothed if you do not choose anything).

Are these weights as in PSIS paper, where weights are defined as smoothed ratios? Truncation or smoothing should be done only for ratios.

MansMeg:
That sounds like a good idea. I suggest we go through this tomorrow - if you have time. 

When we call the psis.matrix() for example. The log_ratios are truncated (if TIS is used). Then the truncated log_ratios are stored in a psis object - after the maximum value has been subtracted from all log_ratios. psis() only returns the log_ratios for *one* IS method, either PSIS, TIS or IS. I have not touched the PSIS method at all, and TIS truncates based on the original log_ratios. 

So a short answer regarding why not call it `tis()`:
I actually started to do it this way, but that would have become a maintenance problem since I was actually copying a lot of code from `psis()`. Now I essentially only add an argument and two very specific functions. 

The downside: psis() also do tis and is - a semantic and clarity problem. 
The upside: Better maintainable and robust code.

My guess is that maintainability trumps semantic problems - since these can be fixed by refactoring and renaming in a later stage.

paul-buerkner:
My quick unsolicited thoughts on that maintainability vs. naming problem. Personally, I would prefer having separate `tis` and `is` methods (although the latter name is already taken by base R), and to refactor the psis code as part of this PR to avoid code duplication. The problem with the adhoc approach of using psis all the way through and then refactoring later, is that somebody has to do the refactoring later, and chances are this will never really happen. loo is one of the core packages of the stanverse in R and we should avoid leaving it in an unwanted temporary stage, at least the master branch and CRAN release. Anyway, just some thoughts of mine.

MansMeg:
Indeed. It could be factored right away. Any suggestions on what to call the generic function that can do any IS approach?

paul-buerkner:
I have to think of that. One way would be to use `tis` with truncation turned off (I assume there will be a `truncation` argument or something in `tis` that specifies the used truncation method). This of course also goes into the direction of overloading a function in a direction that its name does not suggest, but I think would be much less harmful than doing the same thing with `psis` which is arguable way more specific. Writing IS in caps would also be an options, but not consistent with the other function names so probably not desirable.

jgabry:
I think refactoring now to enable general IS is probably a good idea. I don’t have a great idea for the name (it’s annoying that`is` is taken by base R, and we definitely shouldn’t rely on users to do loo::is), but I’ll think about it. The name can always come after the implementation.

MansMeg:
Alright! 

So here is a suggestion. I create the functions `tis()` (truncated is) and `sis()` (standard/simple is) and all functions (including `psis()`) call the general `importance_sampling(..., method)` but with different arguments for method. So the new `psis()` would essentially be:
```
psis <- function(...){
  importance_sampling(..., method = "PSIS)
}
```
The only downside would be that the arguments for PSIS would not be spelled out in the psis function? It can be solved by just @seealso importance_sampling in the documentation, but that is less clear from a user's perspective. I could explicitly write the argument in psis/tis/sis functions, but this would create some additional maintenance burden, but a slightly cleaner API. 

At the same time, I could make sure that `importance_sampling` also doesn't shift the log-weights, as was mentioned in #112?

paul-buerkner:
Duplicating the arguments increases the maintenence burden slightly, as you say, but may be worth at least for the main arguments which users are likely to use frequently.

MansMeg:
Alright! I'll put it together and see what you think.

jgabry:
Is this PR still active or just the other one? If not let's close it

MansMeg:
This PR is an earlier implementation of subsampling loo with tis, but without refactoring. If you think you will merge the refactored PR, then you can close this one. I did not want to assume you wanted half the package refactored. =)

jgabry:
I see, thanks! Let's leave it open until we officially decide on the other one, but I'm leaning towards going with the refactored version in the other PR.

jgabry:
Gonna close this now since we’ll merge your other PR

PR #125, opened 2019-10-18
TIS and SIS implementation using tis() and sis()

MansMeg:
Hi all,

Finally, I got a version done by refactorizing psis to use importance_sampling as a parent class and made psis, tis and sis children of that class. This worked as suggested in our discussion in the previous PR, but required quite a lot additional refactoring. I open up a new PR with this suggestion on implementation so if it does not work I can return to the previous PR by closing this one. I'm not sure if that the correct way. Although, this PR contains everything from the former PR as well.

So the two main discussions  are:
1. Is `importance_sampling` a reasonable function and parent class name for psis, tis and sis?
2. The only place I could not refactorize without changing the API is the loo function. Since we call it loo, rather than psis_loo, I had to add an argument `is_method` if you want to use truncated or standard importance sampling. `psis` is of course default. An alternative would be to have tis_loo(), sis_loo(), psis_loo() etc.

Also, I think I can fix the shifting issue in the issue tracker, but this is not a part of this PR. 

The documentation has not been fixed in this PR, but will be fixed as soon as you are ok with this implementation and before merging.

So what do you think @jgabry , @avehtari @paul-buerkner ?


paul-buerkner:
Some quick thoughts.

1. I am fine with the name but I am certainly not the tree of us with the most experience with the loo package.

2. I think that having an additional argument for loo is not a problem as still the name correctly says what it does regardless of which importance sampling technique is used.

R/importance_sampling.R
```diff
@@ -0,0 +1,209 @@
...
+
+#' @rdname psis
+importance_sampling.matrix <-
+  function(log_ratios,
+           ...,
+           r_eff = NULL,
+           cores = getOption("mc.cores", 1),
+           is_method) {
+    cores <- loo_cores(cores)
+    stopifnot(is_method %in% implemented_is_methods())
```
paul-buerkner: please add an informative error message providing the type of supported IS methods. Right now, I just get, for instance,

```R
Error in importance_sampling.matrix(log_ratios = -x, r_eff = r_eff, cores = cores,  : 
  is_method %in% implemented_is_methods() is not TRUE 
```

which is unhelpful.

R/importance_sampling.R
```diff
@@ -0,0 +1,209 @@
+
+#' @rdname psis
+importance_sampling <- function(log_ratios, ...) UseMethod("importance_sampling")
+
+#' @noRd
+#' @description
+#' Currently implemented importance sampling methods
+implemented_is_methods <- function() c("PSIS", "TIS", "SIS")
```
paul-buerkner: I would argue to use lower case version of the names as the function are called that way. In addition, I would actually argue to specify `is_method = c("psis", "tis", "sis")` in the function signature and then use the canonical `is_method <- match.arg(is_method)` for input validation, which nicely handles strange inputs and provides informative error messages.

R/importance_sampling.R
```diff
@@ -0,0 +1,209 @@
...
+           is_method) {
+    cores <- loo_cores(cores)
+    stopifnot(is_method %in% implemented_is_methods())
+    log_ratios <- validate_ll(log_ratios)
+    r_eff <- prepare_psis_r_eff(r_eff, len = ncol(log_ratios))
+    do_importance_sampling(log_ratios, r_eff = r_eff, cores = cores, is_method = is_method)
+  }
+
+#' @rdname psis
+importance_sampling.default <-
```
paul-buerkner: When I run `loo(fit, is_method = "SIS")` with a brms model for instance and the new loo version, the output is still a `psis_loo` object that tells me that all khat were good although I don't think that SIS actually computed khats at all.

R/loo.R
```diff
@@ -256,20 +258,22 @@ loo.function <-
...
                                     .loo_i = .loo_i,
                                     .llfun = .llfun,
                                     data = data,
                                     draws = draws,
                                     r_eff = r_eff,
                                     save_psis = save_psis,
                                     cores = cores,
-                                    ...)
+                                    ...,
+                                    is_method = is_method)
```
paul-buerkner: is there a reason `is_method` is coming after `...` in the function call?

R/loo_approximate_posterior.R
```diff
@@ -178,9 +178,11 @@ loo_approximate_posterior.function <-
            log_p,
            log_g,
            r_eff = NULL,
-           save_psis = FALSE) {
+           save_psis = FALSE,
+           is_method) {
 
-    if(!is.null(r_eff)) warning("r_eff not implemented for aploo")
+    if(!is.null(r_eff)) warning("r_eff not implemented for aploo.")
```
paul-buerkner: Just a minor thing. Tere and in general, please add spaces after `if` and before `(...)` according to general R code guidelines (same for all special function calls such as `for` etc.) Please also add a space after `(...)` and `{` in `if` and related statements.

R/loo_subsample.R
```diff
@@ -455,7 +455,7 @@ nobs.psis_loo_ss <- function(object, ...){
 #' @param api the choices available in the loo API or all possible choices.
 #' @return a character vector of allowed choices
 loo_approximation_choices <- function(api = TRUE) {
-  lac <- c("plpd", "lpd", "waic", "waic_grad_marginal", "waic_grad", "waic_hess", "none")
+  lac <- c("plpd", "lpd", "waic", "waic_grad_marginal", "waic_grad", "waic_hess", "tis", "none")
```
paul-buerkner: Out of interest: Any reason for not also adding `sis` here?

R/loo.R
```diff
@@ -192,10 +192,10 @@ loo.array <-
            ...,
            r_eff = NULL,
            save_psis = FALSE,
-           cores = getOption("mc.cores", 1)) {
-
+           cores = getOption("mc.cores", 1),
+           is_method = "PSIS") {
```
paul-buerkner: make sure that you propagate `sis` and `tis` to the `loo` class. Right now all `loo` objects are of class psis_loo and loo with associated print method. Please also add proper print methods for `tis_loo` and `sis_loo`.

R/importance_sampling.R
```diff
@@ -0,0 +1,209 @@
+
+#' @rdname psis
+importance_sampling <- function(log_ratios, ...) UseMethod("importance_sampling")
```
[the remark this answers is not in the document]
MansMeg: I'm not sure it should actually be exported? Do we want people to access importance_sampling() directly? I guess we want them to use psis() or tis() or iterative_psis()?

R/importance_sampling.R
```diff
@@ -0,0 +1,209 @@
...
+    stopifnot(length(dim(log_ratios)) == 3)
+    stopifnot(is_method %in% implemented_is_methods())
+    log_ratios <- validate_ll(log_ratios)
+    log_ratios <- llarray_to_matrix(log_ratios)
+    r_eff <- prepare_psis_r_eff(r_eff, len = ncol(log_ratios))
+    do_importance_sampling(log_ratios, r_eff = r_eff, cores = cores, is_method = is_method)
+  }
+
+#' @rdname psis
+importance_sampling.matrix <-
```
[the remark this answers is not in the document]
MansMeg: Even if the method is internal?

R/importance_sampling.R
```diff
@@ -0,0 +1,209 @@
...
+    cores <- loo_cores(cores)
+    stopifnot(length(dim(log_ratios)) == 3)
+    stopifnot(is_method %in% implemented_is_methods())
+    log_ratios <- validate_ll(log_ratios)
+    log_ratios <- llarray_to_matrix(log_ratios)
+    r_eff <- prepare_psis_r_eff(r_eff, len = ncol(log_ratios))
+    do_importance_sampling(log_ratios, r_eff = r_eff, cores = cores, is_method = is_method)
+  }
+
+#' @rdname psis
```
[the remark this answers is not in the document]
MansMeg: Indeed, the question is if importance_sampling should be part of the API or not.

R/importance_sampling.R
```diff
@@ -0,0 +1,209 @@
...
+#' @description
+#' Currently implemented importance sampling methods
+implemented_is_methods <- function() c("PSIS", "TIS", "SIS")
+
+#' @rdname psis
+importance_sampling.array <-
+  function(log_ratios, ...,
+           r_eff = NULL,
+           cores = getOption("mc.cores", 1),
+           is_method) {
```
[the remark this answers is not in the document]
MansMeg: Indeed. I agree!

R/loo.R
MansMeg: No. This could be fixed. You think ... always should be in the end?

R/importance_sampling.R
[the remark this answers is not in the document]
paul-buerkner: Not necessarily, but it appears in the documentation of psis even though users cannot access it directly. If it is not exported, it should not appear in the doc.

R/loo_approximate_posterior.R
MansMeg: This should be something we include in a pre-commit (that I now learnt from Eero) and include in the project. Then it is easily enforced in the test suites as well. I could add this as an issue.

R/importance_sampling.R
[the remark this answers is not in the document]
paul-buerkner: Yes. @export will not export the methods themselves if the generic is not exported but it correctly registers them as S3 methods in the Namespace which is desireble.

R/importance_sampling.R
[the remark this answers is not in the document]
paul-buerkner: It does not need to be but its interfaces, psis, tis, sis should have separate help pages from each other.

R/loo_subsample.R
MansMeg: No, it is just that it has not been studied in our paper. I can include it.

R/loo.R
paul-buerkner: It does not need to be at the end but there should be clear reasons for not having it at the end, which i don't see in this case.

R/loo.R
MansMeg: Indeed! This should be fixed.

R/importance_sampling.R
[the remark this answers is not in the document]
MansMeg: Ok!

R/importance_sampling.R
MansMeg: Indeed!

R/importance_sampling.R
MansMeg: Alright! Sounds good. I'm used to use checkmate that solves this for you. =)

R/importance_sampling.R
MansMeg: Agree. this is a bug. Will fix.

MansMeg:
Alright! I'll fix most thinks in Paul's PR. Although, one question remains. Should the parent constructor `importance_sampling` be exported to users? 

I actually think not in the first step. Simply because we want to add as little as possible to the API. It is always easier to add stuff to the API than to remove it. So we can, in a later stage and if someone demands it, export the parent class, but until then keep it to be internal?

paul-buerkner:
I would agree to keep `importance_sampling` internal.

R/importance_sampling.R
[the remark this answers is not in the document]
MansMeg: Now fixed

R/importance_sampling.R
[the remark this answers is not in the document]
MansMeg: Now fixed

R/importance_sampling.R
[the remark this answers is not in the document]
MansMeg: Fixed

R/loo_subsample.R
MansMeg: Now included

R/loo.R
MansMeg: Now fixed

R/importance_sampling.R
MansMeg: Fixed

R/loo_approximate_posterior.R
MansMeg: Now fixed for all code

R/importance_sampling.R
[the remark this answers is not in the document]
MansMeg: Now fixed

MansMeg:
Alright @paul-buerkner , now everything has been fixed. Please look at it again and then see if there is still any problems? I refactored psis_loo in a similar way as psis - for, of course, consistency. =P

R/importance_sampling.R
```diff
@@ -0,0 +1,219 @@
...
+#' @template is_method
+#' @keywords internal
+importance_sampling <- function(log_ratios, method, ...) UseMethod("importance_sampling")
+
+#' @noRd
+#' @keywords internal
+#' @description
+#' Currently implemented importance sampling methods
+assert_is_method_is_implemented <- function(x){
+  if (!x %in% implemented_is_methods()) stop("Importance sampling method '", x, "' is not implemented. Implemented methods: '", paste0(implemented_is_methods, collapse = "', '"), "'")
```
paul-buerkner: Can you make a line break so that the line does not exceed 80ish characters?

R/importance_sampling.R
```diff
@@ -0,0 +1,219 @@
...
+#' @inheritParams psis
+#' @template is_method
+#' @keywords internal
+importance_sampling <- function(log_ratios, method, ...) UseMethod("importance_sampling")
+
+#' @noRd
+#' @keywords internal
+#' @description
+#' Currently implemented importance sampling methods
+assert_is_method_is_implemented <- function(x){
```
paul-buerkner: using `is` for two different things in the function name is perhaps suboptimal. It is ok for me just perhaps confusing to people reading the code for the first time.

R/loo_approximate_posterior.R
```diff
@@ -153,11 +153,12 @@ loo_approximate_posterior.function <-
...
     }
 
-    ap_psis <- psis_loo_object(
+    ap_psis <- importance_sampling_loo_object(
       pointwise = do.call(rbind, pointwise),
       diagnostics = diagnostics,
       dims = c(attr(psis_list[[1]], "S"), N),
-      psis_object = if (save_psis) psis_out else NULL
+      is_method = "psis",
+      is_object = if (save_psis) psis_out else NULL
```
paul-buerkner: is the name `save_psis` here intentional if we not necessarily have psis objects anymore?

R/sis.R
```diff
@@ -0,0 +1,150 @@
...
+  function(log_ratios, ..., r_eff = NULL) {
+    importance_sampling.default(log_ratios = log_ratios, ...,
+                                r_eff = r_eff,
+                                method = "sis")
+  }
+
+#' @rdname psis
+#' @export
+is.sis <- function(x) {
+  inherits(x, "sis") && is.list(x)
```
paul-buerkner: What is the reason behind using `is.list` here?

R/sis.R
```diff
@@ -0,0 +1,150 @@
...
+#' @noRd
+#' @param log_ratios_i A vector of log importance ratios (for `loo()`, negative
+#'   log likelihoods).
+#' @param tail_len_i Not used. Included to conform to PSIS API.
+#'
+#' @details Implementation standard importance sampling.
+#' @return A named list containing:
+#' * `lw`: vector of unnormalized log weights
+#' * `pareto_k`: scalar Pareto k estimate. For IS, this defaults to 0.
+do_sis_i <- function(log_ratios_i, tail_len_i) {
```
paul-buerkner: Where is `tail_len_i` used in this function? I don't see it here.

R/tis.R
```diff
@@ -0,0 +1,161 @@
...
+  function(log_ratios, ..., r_eff = NULL) {
+    importance_sampling.default(log_ratios = log_ratios, ...,
+                                r_eff = r_eff, method = "tis")
+  }
+
+
+#' @rdname psis
+#' @export
+is.tis <- function(x) {
+  inherits(x, "tis") && is.list(x)
```
paul-buerkner: Same question about `is.list` as above.

R/tis.R
```diff
@@ -0,0 +1,161 @@
...
+#' @details Implementation of Truncated importance sampling (TIS), a method for
+#' stabilizing importance ratios. The version of TIS implemented here
+#' corresponds to the algorithm presented in Ionides (2008) with truncation at
+#' sqrt(S).
+#'
+#' @return A named list containing:
+#' * `lw`: vector of unnormalized log weights
+#' * `pareto_k`: scalar Pareto k estimate. For 'tis', this defaults  to 0.
+#'
+do_tis_i <- function(log_ratios_i, tail_len_i) {
```
paul-buerkner: `tail_len_i` seems to be unused here.

tests/testthat/test_loo_subsampling.R
```diff
@@ -383,6 +383,10 @@ test_that("elpd_loo_approximation works as expected", {
   expect_true(all(pi_vals > pi_vals_waic))
   expect_true(sum(pi_vals) - sum(pi_vals_waic) < 1)
 
+  # Compute tis approximation
+  expect_silent(pi_vals_tis <- loo:::elpd_loo_approximation(.llfun = llfun_test, data = fake_data, draws = fake_posterior, loo_approximation = "tis", loo_approximation_draws = 100, cores = 1))
```
paul-buerkner: break this line to 80ish characters

tests/testthat/test_loo_subsampling.R
```diff
@@ -1063,3 +1067,46 @@ test_that("loo_compare_subsample", {
...
+  y <- rbinom(N, size = K, prob = p)
+  a <- a0 + sum(y); b <- b0 + N * K - sum(y)
+  fake_posterior <- draws <- as.matrix(rbeta(S, a, b))
+  fake_data <- data.frame(y,K)
+  rm(N, K, S, a0, b0, p, y, a, b)
+  llfun_test <- function(data_i, draws) {
+    dbinom(data_i$y, size = data_i$K, prob = draws, log = TRUE)
+  }
+
+  expect_silent(loo_ss_full <- loo_subsample(x = llfun_test, draws = fake_posterior, data = fake_data, observations = 1000, loo_approximation = "plpd", r_eff = rep(1, nrow(fake_data))))
```
paul-buerkner: add some more line breaks here.

R/loo.R
```diff
@@ -215,16 +218,18 @@ loo.matrix <-
...
            r_eff = NULL,
            save_psis = FALSE,
-           cores = getOption("mc.cores", 1)) {
-
+           cores = getOption("mc.cores", 1),
+           is_method = c("psis", "tis", "sis")) {
     if (is.null(r_eff)) throw_loo_r_eff_warning()
-    psis_out <- psis.matrix(log_ratios = -x, r_eff = r_eff, cores = cores)
+    is_method <- match.arg(is_method)
+    psis_out <- importance_sampling.matrix(log_ratios = -x, r_eff = r_eff, cores = cores, method = is_method)
```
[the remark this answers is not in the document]
MansMeg: So this is a more general question. I first started to call it is_object. But then that would have meant that I would have changed the API argument `save_psis`, and changing that would have been a large (even not back compatible) change in the API. So this can of course be done, but the question is then what we should do with the save_psis argument. Now I did as little as possible changes.

R/loo.R
```diff
@@ -215,16 +218,18 @@ loo.matrix <-
            ...,
            r_eff = NULL,
            save_psis = FALSE,
-           cores = getOption("mc.cores", 1)) {
-
+           cores = getOption("mc.cores", 1),
+           is_method = c("psis", "tis", "sis")) {
```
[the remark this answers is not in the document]
MansMeg: So it has been changed to method in `importance_sampling()` constructor since there `method` makes sense (it is the importance sampling method used). But in loo this becomes more complicated.  Using method is semantically slightly difference since there it would be the loo method. To me that could be a naive approach, approximation using Taylor series (as in waic)    etc. So in loo I kept `is_method` to make it more explicit. Maybe call it `importance_sampling_method` but that may be too long?

R/loo.R
```diff
@@ -256,20 +263,22 @@ loo.function <-
...
                                     .loo_i = .loo_i,
                                     .llfun = .llfun,
                                     data = data,
                                     draws = draws,
                                     r_eff = r_eff,
                                     save_psis = save_psis,
                                     cores = cores,
-                                    ...)
+                                    ...,
+                                    method = is_method)
```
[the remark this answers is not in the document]
MansMeg: No. I thought I fixed all these. Now it has been fixed.

R/importance_sampling.R
MansMeg: Indeed. Ill make this more clear.

R/sis.R
MansMeg: It is not used. It comes from the fact that `tail_len_i` is supplied by default to any `do_XXis_i` function. Maybe state that this as a comment in the code?

R/tis.R
MansMeg: This a copy from the `is.psis()` function that has that exact structure.

R/loo.R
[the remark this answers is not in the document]
paul-buerkner: yeah, I get it. Let's keep it `save_psis` for now, which we may as well eventually deprecate in favor of another name. But I agree this would be too much for now.

R/loo.R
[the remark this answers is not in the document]
paul-buerkner: Ah I see. Ok, from this perspective `is_method` is fine for me.

R/sis.R
paul-buerkner: Can we replace it by `...` so that anything may be passed but is then simply ignored? This may be a cleaner approach than having an argument which actually takes no effect.

R/tis.R
paul-buerkner: Ok. Fair enough. Let's keep it that way for now as it didn't cause any trouble in the past.

R/sis.R
MansMeg: I fully agree. I'll see if that solves the problem.

MansMeg: Seem to work.

MansMeg:
Alright. Now, this has been fixed @paul-buerkner . Thanks for the comments! I'm really home blind in the code now. =)

paul-buerkner:
I am happy now! Thanks for putting so much effort into this!

@jgabry could you please take a look and merge the PR if you think it is ready?

jgabry:
I’ve been super busy but am coming back to this now. Since there’s been a lot of discussion on this in various comments in different places, would one of you be willing to put together a list (or summary of some kind) of differences between the old way and this new way? That would help me when reviewing and also could be useful to have such a summary to draw from for documentation purposes. Thanks and sorry for the delay on this! So much stuff going on :)

jgabry:
And thanks for all the work on this so far!!

MansMeg:
Sure!

So this comes from that I implemented truncated importance sampling for subsampling purposes and Aki though that would be good to have in loo as well, as an alternative to psis. To do it in a good way I refactored the classes to accommodate this. This means that:

I added an `importance_sampling` parent class. So `"psis"` -> `"psis", "importance_sampling"` and  `"psis_loo", "loo"` -> `"psis_loo", "importance_sampling_loo", "loo"`.

Then I fixed so all functions that are common to all IS objects are using the "importance_sampling" methods, such as dim() and weight(). In this way, we can add any new IS technique relatively simple, such as the moment matching approach.

The only change of the  API from the users' perspective is that loo now has the argument `is_method` that defaults to `"psis"`. Also, the functions tis() and sis() are added to use truncated and standard importance sampling.

I hope this summarizes it.

jgabry:
Thanks!

jgabry:
1) At first glance this looks great! Thanks for doing all this @MansMeg and @paul-buerkner for reviewing. 
2) I'll look it over in more detail within the next few days and maybe push some edits to the doc. (if I have access to your repo, or I think you have access to loo so maybe you could just do this from a branch off of loo or make this PR to a branch of loo that I'll merge immediately and then we'll go from there to master, if that makes sense.)  
3) I think this is enough new stuff to warrant a new vignette or mention in an existing vignette if there's a spot that makes sense. This may never happen unless someone does it now, if you know what I mean ;)

MansMeg:
2. I think the easiest is for you to just push to my fork? I'm not sure if we would lose this PR discussion otherwise? I'll give you access to it.
3. I guess most things are just changes in the backend and should not change at all for the users? So maybe just add the "is_method" in one of the ordinary vignette. Although I do not think we would want any users to use tis() or sis() for loo anyway? Only as an approximation in large data psis-loo?

jgabry:
For (2) that's fine! 

For (3) yeah I agree we don't want anyone using tis() or sis() for loo, I was just thinking that other people might want to use them for research. But a whole new vignette probably isn't necessary

jgabry:
I haven’t forgotten about this, just been super busy. I’ve started going through this again and it looks good. Will finish asap

MansMeg:
Great!

jgabry:
Looks like Travis is failing because of doc warnings. I’ll fix and then merge!

PR #130, opened 2019-11-20
Add mmloo and split_mmloo

topipa:
Hi!

Here is the pull request for the moment matching functionality.

New functionality:
1. function `mmloo`, which takes in a `loo` object and modifies it to give more accurate estimates. This requires functions for recomputing posterior densities and likelihoods, which need to be implemented for example in Rstan and other packages when they use mmloo.
2. function `split_mmloo`, which `mmloo` uses internally if its parameter `split = TRUE`. Using the split gives more accurate estimates and is thus true by default.

Changes made to existing functions/files:
3. the `pointwise` field now also contains pareto k diagnostics indicating the influence of each observation, which are the same as in the `diagnostics` field. These are not summarized in `estimates` and are never modified by mmloo. mmloo only modifies the pareto k's in `diagnostics`.
4. I rewrote the documentation of loo and pareto-k-diagnostic to reflect the changes. Feel free to comment and suggest improvement to the documentation.

Questions/points to consider:
5. how to name the new leverage pareto k diagnostic? Currently it is `leverage_pareto_k`.
6. Currently MCMC relative effective sample size is recomputed after the moment matching, even though the draws are no longer strictly from an MCMC chain. For very influential observations this seems to improve the MCMC effective sample size. Do you think this is sensible or not? Additionally, when using `split_mmloo`, half of the draws come from the original full data posterior, and half are transformed by moment matching. How to compute relative_eff in this case? Currently it is computed separately from both, and the smaller of those is used.
7. I have used integrand-specific IS effective sample size in mmloo, because otherwise the IS effective sample size may be heavily overestimated (https://statweb.stanford.edu/~owen/mc/Ch-var-is.pdf eq. 9.17). After mmloo, the IS effective sample size is the minimum of the _normal IS effective sample size_ and _integrand-specific IS effective sample size_.
8. As an example, I implemented one helper function for extracting the `leverage_pareto_k` values. We can decide whether these are needed or not.
9. I have made some tests, but have not done any profiling or code coverage checks.
10. I have not bumped the DESCRIPTION file.


jgabry:
Awesome, thanks @topipa! I’ll try to go through this and respond to your questions/points to consider soon (it might be next week at this point as I’m traveling tomorrow). 

Also, @avehtari @paul-buerkner @MansMeg @yao-yl feel free to chime in if you have any thoughts on the implementation or the questions/points to consider and/or want to help review this.

paul-buerkner:
Thanks Jonah! I have contributed to the code so I should not be the only one looking at this PR but I will review and test it anyway.

MansMeg:
Sure! @topipa . Your PR seem to fail? Have you checked what crashes?

topipa:
I think I forgot to run the full test suite after the last modifications. I will check that!

topipa:
Okay, the tests are passing now after small modifications to loo_subsample.R . Could you @MansMeg check that the changes are reasonable?

paul-buerkner:
In the new mmloo brms branch, I implement the mmloo method in brms based on Topi's loo branch. This seems to work well, so the PR passes the first hurdle for me. I will start reviewing the actual code later on today.

R/loo_moment_matching.R
```diff
@@ -0,0 +1,565 @@
...
+  UseMethod("mmloo")
+}
+
+
+#' @describeIn mmloo A default method that takes as arguments
+#' a user-specified model object \code{x}, a \code{loo} object and
+#' user-specified functions \code{post_draws}, \code{log_lik},
+#' \code{unconstrain_pars}, \code{log_prob_upars}, and \code{log_lik_upars}.
+#' @export
+mmloo.default <- function(x, loo, post_draws, log_lik,
```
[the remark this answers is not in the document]
paul-buerkner: Should we perhaps rename `log_lik` and `log_lik_upars` to `log_lik_i` and `log_lik_upars_i` to make clear it is specifically to compute log-likelihood values of a single observation?

R/loo_moment_matching.R
```diff
@@ -0,0 +1,568 @@
...
+  checkmate::assertFunction(log_prob_upars)
+  checkmate::assertFunction(log_lik_upars)
+  checkmate::assertNumber(max_iters)
+  checkmate::assertNumber(k_thres)
+  checkmate::assertLogical(split)
+  checkmate::assertLogical(cov)
+  checkmate::assertNumber(cores)
+
+
+  if ("psis_loo" %in% class(loo)) {
```
paul-buerkner: Does it make sense to clean up the commented out code below?

R/loo_moment_matching.R
```diff
@@ -0,0 +1,568 @@
...
+    # pointwise estimates
+    # elpd_loo
+    elpd_loo_i <- matrixStats::logSumExp(log_liki + lwi)
+    loo$pointwise[i, "elpd_loo"] <- elpd_loo_i
+    # p_loo
+    lpd <- matrixStats::logSumExp(log_liki) - log(length(log_liki))
+    loo$pointwise[i, "p_loo"] <- lpd - elpd_loo_i
+    # mcse_elpd_loo
+    loo$pointwise[i, "mcse_elpd_loo"] <- mcse_elpd(as.matrix(log_liki),as.matrix(lwi),
+                                     exp(elpd_loo_i), r_effi)
```
paul-buerkner: Perhaps write this as 

```R
loo$pointwise[i, "mcse_elpd_loo"] <- mcse_elpd(
  as.matrix(log_liki), as.matrix(lwi), exp(elpd_loo_i), r_effi
)
```

with the arguments names made explit (as we currently have four unnamed arguments here)?

R/loo_moment_matching.R
```diff
@@ -0,0 +1,568 @@
...
+    was reached.\n",
+    "Increasing the value may improve accuracy.",
+    call. = FALSE
+  )
+}
+
+#' Warning message if not using split transformation and accuracy is compromised
+#' @noRd
+throw_large_kf_warning <- function(kf) {
+  if (any(kf > 0.5)) {
```
paul-buerkner: does it make sense to make the threshold an argument defaulting to 0.5?

R/split_moment_matching.R
```diff
@@ -0,0 +1,128 @@
+#' Split moment matching for efficient approximate leave-one-out cross-validation (LOO)
+#'
+#' A function that computes the split moment matching importance sampling loo.
+#' Takes in the moment matching total transformation, transforms only half
+#' of the draws, and computes a single elpd using multiple importance sampling.
+#'
+#'
+#' @export split_mmloo
```
paul-buerkner: export statements do not need to be more specific if they appear in the doc of a function that should be exported via this statement.

paul-buerkner: do we need to export split_mmloo at this point or shall we keep it internal and only make it callable via mmloo?

tests/testthat/test_loo_moment_matching.R
```diff
@@ -0,0 +1,303 @@
+library(loo)
+library(rstan)
+options(mc.cores = 1)
+
+
+context("moment matching")
+set.seed(123)
+S <- 40000
```
paul-buerkner: do we need that many posterior draws?

tests/testthat/test_loo_moment_matching.R
```diff
@@ -0,0 +1,303 @@
...
+
+
+
+
+
+# implement functions for moment matching loo
+
+
+# extract original posterior draws
+post_draws_test <- function(x) {
```
paul-buerkner: all of these functions should contain a `...` argument even if not actually used as otherwise they will break if any kind of additional argument is passed to one of these functions.

R/loo_moment_matching.R
```diff
@@ -0,0 +1,568 @@
...
+#'   If \code{FALSE}, only the mean and marginal variances are matched.
+#' @template cores
+#' @param ... Further arguments passed to the custom functions documented above.
+#'
+#' @return The `mmloo()` methods return an updated \code{loo} object.
+#' The structure of the updated \code{loo} object is similar, but the
+#' method also stores the original Pareto k diagnostic values in
+#' the diagnostics field.
+#'
+#' @details The `mmloo()` function is an S3 generic and we provide a default
```
paul-buerkner: We should indicate here that all of these functions should contain a `...` argument even if not actually used as otherwise they will break if any kind of additional argument is passed to one of these functions.

tests/testthat/test_loo_moment_matching.R
```diff
@@ -0,0 +1,303 @@
...
+
+# mu, log(sigma)
+unconstrain_pars_test <- function(x, pars) {
+  upars <- as.matrix(pars)
+  upars[,2] <- log(upars[,2])
+  upars
+}
+
+log_prob_upars_test <- function(x, upars) {
+  dinvchisq(exp(upars[,2])^2,x$data$n - 1,x$data$s2, log = TRUE) + dnorm(upars[,1],x$data$ymean,exp(upars[,2])/sqrt(x$data$n), log = TRUE)
```
paul-buerkner: add a line break here

tests/testthat/test_loo_moment_matching.R
```diff
@@ -0,0 +1,303 @@
+library(loo)
```
paul-buerkner: do we have tests somehwere that the diagnostic pareto k values are actually updated while the leverage values remain the same?

paul-buerkner:
I think the PR already looks quite good and I hope my comments help to further improve it.

R/loo_moment_matching.R
```diff
@@ -0,0 +1,565 @@
...
+
+#' @describeIn mmloo A default method that takes as arguments
+#' a user-specified model object \code{x}, a \code{loo} object and
+#' user-specified functions \code{post_draws}, \code{log_lik},
+#' \code{unconstrain_pars}, \code{log_prob_upars}, and \code{log_lik_upars}.
+#' @export
+mmloo.default <- function(x, loo, post_draws, log_lik,
+                          unconstrain_pars, log_prob_upars,
+                          log_lik_upars, max_iters = 30L,
+                          k_thres = 0.5, split = TRUE,
```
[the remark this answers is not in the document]
avehtari: I would go with 0.7 unless there is some evidence I don't remember for 0.5

R/diagnostics.R
```diff
@@ -143,6 +162,27 @@ pareto_k_values <- function(x) {
...
 }
 
+#' @rdname pareto-k-diagnostic
+#' @export
+#' @return `pareto_k_leverage_values()` returns a vector of the estimated Pareto
+#'   \eqn{k} parameters. These represent influence of the observations on the
+#'   model posterior distribution.
+pareto_k_leverage_values <- function(x) {
+  k <- x$pointwise[,"leverage_pareto_k"]
+  if (is.null(k)) {
```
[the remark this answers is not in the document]
topipa: Thanks! Fixed already!

R/loo.R
```diff
@@ -220,17 +228,9 @@ loo.matrix <-
...
-      throw_loo_r_eff_warning()
-    }
-    psis_out <-
-      importance_sampling.matrix(
-        log_ratios = -x,
-        r_eff = r_eff,
-        cores = cores,
-        method = is_method
-      )
+    psis_out <- importance_sampling.matrix(log_ratios = -x, r_eff = r_eff, cores = cores, method = is_method)
```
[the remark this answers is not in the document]
topipa: no reason

R/loo_moment_matching.R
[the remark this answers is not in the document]
topipa: Fixed

R/loo.R
```diff
@@ -271,19 +271,17 @@ loo.function <-
...
+                                    N = N,
+                                    .loo_i = .loo_i,
+                                    .llfun = .llfun,
+                                    data = data,
+                                    draws = draws,
+                                    r_eff = r_eff,
+                                    save_psis = save_psis,
+                                    cores = cores,
+                                    method = is_method,
+                                    ...)
```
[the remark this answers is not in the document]
topipa: no reason

R/loo_moment_matching.R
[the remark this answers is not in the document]
topipa: I think `k_threshold` is better. 4 characters is insignificant for the amount of clarity lost. Changed to `k_threshold` unless someone has other opinions.

R/loo_moment_matching.R
topipa: cleaned

R/loo_moment_matching.R
topipa: done

R/loo_moment_matching.R
topipa: Changed

R/split_moment_matching.R
topipa: Just to be clear, you mean that we can simply put "@export" ?

topipa: That is probably better. I will do that unless someone has a different opinion

tests/testthat/test_loo_moment_matching.R
topipa: fixed

R/loo_moment_matching.R
topipa: Done!

tests/testthat/test_loo_moment_matching.R
topipa: This happens in `test_that("mmloo.default works", {`. I added an explicit tests to check this.

paul-buerkner:
Yes just use export.

Topi Paananen <notifications@github.com> schrieb am Fr., 22. Nov. 2019,
09:49:

tests/testthat/test_loo_moment_matching.R
topipa: Maybe that's a bit excessive :) Reduced to 4000.

topipa:
Changes suggested by @paul-buerkner are now done! Could you take another look and see if they look okay?

topipa:
For some reason, the latest commit is not passing in AppVeyor even though it only changed the names of the log_lik and log_lik_uaprs functions. From the AppVeyor output, I could not figure out why it is failing..

MansMeg:
Do you want me to do a code review still or has that already been done by @paul-buerkner ?

paul-buerkner:
I already made one round and will do a second today at some point.

MansMeg:
Ok! Great, then I skip doing a review. I just saw now that it is Aki and you that are assigned to review.

jgabry:
Sorry for the delay on this. I'm finally mostly recovered from the flu!

CRAN requested a loo release in the next few days to fix the error at https://cran.r-project.org/web/checks/check_results_loo.html, so I will take care of that and release what's currently on master, and then we'll get this PR merged after that.

R/split_moment_matching.R
```diff
@@ -0,0 +1,127 @@
...
+#' and the integrand-specific log-importance weights.
+#'
+#'
+#' @seealso [loo()], [mmloo()]
+#' @template moment-matching-references
+#'
+#'
+split_mmloo <- function(x, upars, cov, total_shift, total_scaling,
+                     total_mapping, i, log_prob_upars,
+                     log_lik_i_upars, r_effi, cores,
```
jgabry: Do you mind changing the name of the argument from `r_effi` to `r_eff_i` for consistency with other names?

jgabry:
Actually one more thing in addition to changing the `r_effi` argument name: 

The name `mmloo` I think for consistency should be `mm_loo` (like `ap_psis`) or `loo_moment_match` (like `loo_subsample`) or something like that. But `mmloo` doesn't seem to fit our naming conventions. Does that make sense? 

What do you all (@topipa @paul-buerkner @avehtari @MansMeg) prefer?

avehtari:
I'm also thinking if it should be `imm` instead `of `mm`? But I let @topipa to decide that one.

R/split_moment_matching.R
topipa: Thanks! This is now changed

paul-buerkner:
Thank you Jonah for your review!

I am open for another function name as well. While we are at it, let's also discuss the name of the exact loo for problematic observations which I currently call `reloo`. Since `mmloo` (or however we will name it) is basically a more efficienct substitute of `reloo`, I think their naming conventions should also be similar.

topipa:
I am open to changing the name. `mmloo` was chosen to be similar to `reloo` because `mmloo` is a computationally cheaper alternative to `reloo` for improving accuracy.

topipa:
> I'm also thinking if it should be `imm` instead `of `mm`? But I let @topipa to decide that one.

It's hard to say which would be better. I would lean towards omitting the iterative word because it is shorter that way, and there will most likely never be non-iterative moment matching. Also, referring to Paul's comment, it is a nice bonus if `reloo` and `mmloo` have the same number of characters.

paul-buerkner:
I am happy to rename 'reloo' as well of course.  I just think the names
should be consistent in some way.

Topi Paananen <notifications@github.com> schrieb am Do., 6. Feb. 2020,
15:51:

avehtari:
> let's also discuss the name of the exact loo for problematic observations which I currently call reloo

I guess `reloo` comes from doing refits. It could be called also exact loo `exloo` or brute force loo `bfloo`

topipa:
Correct me if I am wrong, but it seems that the function name `reloo` is used in both BRMS and RStanArm. However, only BRMS has `reloo` as a parameter in the `loo` function, and in RStanArm it is implicitly set to true when `k_threshold` is set to something other than `NULL`. I'm not sure what is the origin of the difference, but changing the user interfaces is probably something we don't want to do lightly?

If we talk only about the function names: I like Jonah's suggestion of `loo_moment_match` because it is consistent with `loo_subsample`. For `reloo`, `loo_refit` or something similar would be consistent with these.

paul-buerkner:
I like the name `loo_moment_match` and also `loo_refit`.

avehtari:
I also like longer names as they are more descriptive

jgabry:
> I like the name loo_moment_match and also loo_refit.

I like these too!

paul-buerkner:
Then let's go for them. @topipa can you make the name change to `mmloo`? Then we should be able to merge it. @jgabry will you do the merge once the name change is done?

I will rename `reloo` in brms in the next couple of days. Eventually, `loo_refit` should also be a generic in the loo package following a similar modular structure as `mmloo` making it easy for other package others to have their own `loo_refit` functions. But this is not within the scope of this PR of course.

jgabry:
Sounds good!

MansMeg:
Late into this discussion, I also favor loo_moment_match and loo_refit, clarity always wins the long game. =)

topipa:
Thank you all for giving your opinions! I will now change `mmloo` to `loo_moment_match` and `split_mmloo` to `loo_moment_match_split`

topipa:
In our discussion with @avehtari and @paul-buerkner two additional points came up. 

1. We discussed the naming of `leverage_pareto_k` (which are not changed by moment matching or refitting) and agreed that `influence_pareto_k` may be a more accurate name to describe the influence each observation has on the posterior. I will change that unless someone disagrees.


2. It would be convenient to store information about which algorithm the elpd values come from. Since moment matching and refitting operate on individual observations, the information should be stored separately for each observation. For example, by default this value could be the same as `is_method` for all observations, but moment matching or refitting could change it from, say, `psis` to `refit` or `psis-mm` for some observations. What do you think about this idea @jgabry ?

Details to discuss that come to my head at this moment
- Where to store it? `pointwise` is currently a matrix, could we make it a data frame to include a column with text? It could also be an element in `diagnostics`?
- What kind of information to include and how to name the different options (is method/moment matching/refit)? It seems that even `is_method` is currently not stored anywhere. Probably subsampling would be indicated here as well? @MansMeg

avehtari:
> We discussed the naming of leverage_pareto_k (which are not changed by moment matching or refitting) and agreed that influence_pareto_k may be a more accurate name to describe the influence each observation has on the posterior. I will change that unless someone disagrees.

Pinging @dpsimpson as he wished that we keep Pareto k as an influence measure

dpsimpson:
Influence_pareto_k is a good name. It should be consistent across the whole package though (ie a global swap pareto_k -> influence_k / influence_pareto_k / k_hat_influence

paul-buerkner:
We will keep one pareto-k vector, which may change as we apply methods such a moment matching and another vector (the influence k one), which stays as it is. I am not sure the influence one should be the one displayed in the standard print outputs as these outputs are mostly concerned with the accuracy of LOO itself and less with influential observations. The latter should receive its own new methods to extract it (if we don't already have a method for this already). But this doesn't necessarily need to be part of this PR.

MansMeg:
So without looking in the actual code, I think `is_method` is stored in the matrix (since we could have different methods for different observations). 

The subsampling loo is a little different since it can essentially use any is_algorithm, either as a surrogate or as the subsampled observation. I would for example like to use moment_match_loo on the subsampled observations, but maybe just the training error or TIS for the surrogates?

topipa:
Actually, the discussion about storing the algorithm information goes so much beyond moment matching that I will create a new issue dedicated to it, so that this pull request does not get too cluttered and can be merged.

paul-buerkner:
I think this is a good idea!

Am Do., 13. Feb. 2020 um 12:40 Uhr schrieb Topi Paananen <
notifications@github.com>:

topipa:
The `leverage_pareto_k` is now renamed to `influence_pareto_k`. I think this could be merged now?

jgabry:
Ok great, merging now!!

PR #139, opened 2020-02-20
Vignette for moment matching

topipa:
This PR adds a vignette about moment matching. It shows an example of using moment matching directly with `loo.stanfit` based on my code in https://github.com/topipa/rstan (https://github.com/stan-dev/rstan/pull/743). It also includes an example of calling `loo_moment_match` directly and how the helper functions need to be like.

### DIFF
--- a/vignettes/loo2-moment-matching.Rmd
+++ b/vignettes/loo2-moment-matching.Rmd
@@ -1,0 +1,301 @@
+---
+title: "Avoiding model refits in leave-one-out cross-validation with moment matching"
+author: "Topi Paananen, Paul Bürkner, Aki Vehtari and Jonah Gabry"
+date: "`r Sys.Date()`"
+output:
+  html_vignette:
+    toc: yes
+params:
+  EVAL: !r identical(Sys.getenv("NOT_CRAN"), "true")
+---
+<!--
+%\VignetteEngine{knitr::rmarkdown}
+%\VignetteIndexEntry{Avoiding model refits in leave-one-out cross-validation with moment matching}
+-->
+
+# Introduction
+
+This vignette demonstrates how to improve the Monte Carlo sampling accuracy
+of leave-one-out cross-validation with the __loo__ package and Stan. The
+__loo__ package automatically monitors the sampling accuracy using
+Pareto $k$ diagnostics for each observation. Here, we present a method for quickly improving
+the accuracy, when the Pareto diagnostics indicate problems. The methodology
+presented is based on the paper
+
+* Paananen, T., Piironen, J., Buerkner, P.-C., Vehtari, A. (2019).
+Pushing the Limits of Importance Sampling through Iterative Moment Matching.
+[arXiv preprint arXiv:1906.08850](http://arxiv.org/abs/1906.08850).
+
+More information about the Pareto $k$ diagnostics is given in the following
+papers
+
+* Vehtari, A., Gelman, A., and Gabry, J. (2017). Practical Bayesian model evaluation using leave-one-out cross-validation and WAIC. _Statistics and Computing_. 27(5), 1413--1432. \doi:10.1007/s11222-016-9696-4. Links: [published](http://link.springer.com/article/10.1007\%2Fs11222-016-9696-4) | [arXiv preprint](http://arxiv.org/abs/1507.04544).
+
+* Vehtari, A., Gelman, A., and Gabry, J. (2017). Pareto smoothed importance sampling. [arXiv preprint arXiv:1507.04544](http://arxiv.org/abs/1507.04544).
+
+# Example: Eradication of Roaches
+
+We will use the same example as in the case study
+[_Bayesian data analysis - roaches cross-validation demo_](https://avehtari.github.io/modelselection/roaches.html).
+See the demo for a description of the problem and data. We will use the same Poisson
+regression model as in the case study.
+
+## Coding the Stan model
+
+Here is the Stan code for fitting the Poisson regression model, which 
+we will use for modelling the number of roaches.
+
+```{r, stancode}
+stancode = "
+data {
+  int<lower=1> K;
+  int<lower=1> N;
+  matrix[N,K] x;
+  int y[N];
+  vector[N] offset;
+
+  real beta_prior_scale;
+  real alpha_prior_scale;
+}
+parameters {
+  vector[K] beta;
+  real intercept;
+}
+model {
+  y ~ poisson(exp(x * beta + intercept + offset));
+  beta ~ normal(0,beta_prior_scale);
+  intercept ~ normal(0,alpha_prior_scale);
+}
+generated quantities {
+  vector[N] log_lik;
+  for (n in 1:N)
+    log_lik[n] = poisson_lpmf(y[n] | exp(x[n] * beta + intercept + offset[n]));
+}
+"
+```
+
+Following the usual approach recommended in 
+[_Writing Stan programs for use with the loo package_](http://mc-stan.org/loo/articles/loo2-with-rstan.html), 
+we compute the log-likelihood for each observation in the 
+`generated quantities` block of the Stan program.
+
+
+## Setup
+
+In addition to __loo__, we load the __rstan__ package for fitting the model,
+and the __rstanarm__ package for the data.
+
+```{r, setup, message=FALSE}
+library("rstan")
+library("loo")
+library("rstanarm")
+seed <- 9547
+set.seed(seed)
+```
+
+## Fitting the model with RStan
+
+Next we fit the model in Stan using the __rstan__ package:
+
+```{r, modelfit}
+# Prepare data
+data(roaches)
+roaches$roach1 <- sqrt(roaches$roach1)
+y <- roaches$y
+x <- roaches[,2:4]
+offset <- log(roaches[,5])
+n <- dim(x)[1]
+k <- dim(x)[2]
+
+standata <- list(N = n, K = k, x = as.matrix(x), y = y,offset=offset,beta_prior_scale = 2.5,alpha_prior_scale = 5.0)
+
+# Compile
+stanmodel <- stan_model(model_code = stancode)
+
+# Fit model
+fit <- sampling(stanmodel, data = standata, seed = seed, refresh = 0)
+print(fit, pars = "beta")
+```
+
+Let us now evaluate the predictive performance of the model using `loo()`.
+
+```{r, loo1}
+loo1 <- loo(fit)
+loo1
+```
+
+We notice that the `loo()` output warns us that there are some observations
+which are very influential, and thus the
+accuracy of importance sampling is compromised as indicated by the large Pareto $k$
+diagnostic values (> 0.7). As discussed in the demo
+[_Bayesian data analysis - roaches cross-validation demo_](https://avehtari.github.io/modelselection/roaches.html),
+this may be an indication of model misspecification. Despite that, it is still beneficial
+to be able to evaluate the predictive performance of the model accurately.
+
+## Moment matching correction for importance sampling
+
+To improve the accuracy of the `loo()` result above, we could perform
+leave-one-out cross-validation by explicitly leaving out single observations
+and refitting the model repeatedly. However, the Pareto $k$ diagnostics
+indicate that there are 19 observations which are problematic. This would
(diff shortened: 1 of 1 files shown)


vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,301 @@
...
+
+# Introduction
+
+This vignette demonstrates how to improve the Monte Carlo sampling accuracy
+of leave-one-out cross-validation with the __loo__ package and Stan. The
+__loo__ package automatically monitors the sampling accuracy using
+Pareto $k$ diagnostics for each observation. Here, we present a method for quickly improving
+the accuracy, when the Pareto diagnostics indicate problems. The methodology
+presented is based on the paper
+
```
avehtari: add a comment about two different Pareto $k$'s

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,301 @@
...
+```
+
+Let us now evaluate the predictive performance of the model using `loo()`.
+
+```{r, loo1}
+loo1 <- loo(fit)
+loo1
+```
+
+We notice that the `loo()` output warns us that there are some observations
```
avehtari: Remove "We notice that"

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,301 @@
...
+this may be an indication of model misspecification. Despite that, it is still beneficial
+to be able to evaluate the predictive performance of the model accurately.
+
+## Moment matching correction for importance sampling
+
+To improve the accuracy of the `loo()` result above, we could perform
+leave-one-out cross-validation by explicitly leaving out single observations
+and refitting the model repeatedly. However, the Pareto $k$ diagnostics
+indicate that there are 19 observations which are problematic. This would
+require 19 model refits which requires lots of computation time.
```
avehtari: which can require lot of

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,301 @@
...
+diagnostic values (> 0.7). As discussed in the demo
+[_Bayesian data analysis - roaches cross-validation demo_](https://avehtari.github.io/modelselection/roaches.html),
+this may be an indication of model misspecification. Despite that, it is still beneficial
+to be able to evaluate the predictive performance of the model accurately.
+
+## Moment matching correction for importance sampling
+
+To improve the accuracy of the `loo()` result above, we could perform
+leave-one-out cross-validation by explicitly leaving out single observations
+and refitting the model repeatedly. However, the Pareto $k$ diagnostics
```
avehtari: refitting the model using MCMC repeatedly

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,301 @@
...
+indicate that there are 19 observations which are problematic. This would
+require 19 model refits which requires lots of computation time.
+
+Instead of refitting, we can perform a moment matching correction to
+the importance sampling for the problematic observations, avoiding refitting
+the model. This can be done with the `loo_moment_match()` function from the
```
avehtari: remove ", avoiding refitting the model" (it's already said in the beginning of the sentence)

function from -> function in

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,301 @@
...
+the Pareto $k$ threshold, above which moment matching is used.
+By default, it operates
+on all observations whose Pareto $k$ value is larger than 0.7.
+(NOTE: Using moment matching directly using the `loo()` function is currently
+available in the development version of __rstan__ in
+[https://github.com/topipa/rstan](https://github.com/topipa/rstan), but will
+be available in the CRAN release eventually).
+
+```{r, message=FALSE}
+loo2 <- loo(fit,moment_match = TRUE)
```
avehtari: add space after comma

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,301 @@
...
+
+Stan Development Team (2018) _RStan: the R interface to Stan, Version 2.17.3._   http://mc-stan.org
+
+Paananen, T., Piironen, J., Buerkner, P.-C., Vehtari, A. (2019).
+Pushing the Limits of Importance Sampling through Iterative Moment Matching.
+[arXiv preprint arXiv:1906.08850](http://arxiv.org/abs/1906.08850).
+
+Vehtari, A., Gelman, A., and Gabry, J. (2017). Practical Bayesian model evaluation using leave-one-out cross-validation and WAIC. _Statistics and Computing_. 27(5), 1413--1432. \doi:10.1007/s11222-016-9696-4. Links: [published](http://link.springer.com/article/10.1007\%2Fs11222-016-9696-4) | [arXiv preprint](http://arxiv.org/abs/1507.04544).
+
+Vehtari, A., Gelman, A., and Gabry, J. (2017). Pareto smoothed importance sampling. [arXiv preprint arXiv:1507.04544](http://arxiv.org/abs/1507.04544).
```
avehtari: update the reference to include Simpson and Yao (see the latest arXiv version)

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,301 @@
...
+papers
+
+* Vehtari, A., Gelman, A., and Gabry, J. (2017). Practical Bayesian model evaluation using leave-one-out cross-validation and WAIC. _Statistics and Computing_. 27(5), 1413--1432. \doi:10.1007/s11222-016-9696-4. Links: [published](http://link.springer.com/article/10.1007\%2Fs11222-016-9696-4) | [arXiv preprint](http://arxiv.org/abs/1507.04544).
+
+* Vehtari, A., Gelman, A., and Gabry, J. (2017). Pareto smoothed importance sampling. [arXiv preprint arXiv:1507.04544](http://arxiv.org/abs/1507.04544).
+
+# Example: Eradication of Roaches
+
+We will use the same example as in the case study
+[_Bayesian data analysis - roaches cross-validation demo_](https://avehtari.github.io/modelselection/roaches.html).
```
avehtari: cite loo2-example.html in loo vignettes instead

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,309 @@
...
+
+## Setup
+
+In addition to __loo__, we load the __rstan__ package for fitting the model,
+and the __rstanarm__ package for the data.
+
+```{r, setup, message=FALSE}
+library("rstan")
+library("loo")
+library("rstanarm")
```
paul-buerkner: If we just need the data from rstanarm, we shouldn't load the package and instead load the data as `data(roaches, package = "rstanarm")`

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,309 @@
...
+## Fitting the model with RStan
+
+Next we fit the model in Stan using the __rstan__ package:
+
+```{r, modelfit}
+# Prepare data
+data(roaches)
+roaches$roach1 <- sqrt(roaches$roach1)
+y <- roaches$y
+x <- roaches[,2:4]
```
paul-buerkner: selecting variables via column indices is not very descriptive. Can we use explicit variables names?

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,309 @@
...
+```
+
+## Fitting the model with RStan
+
+Next we fit the model in Stan using the __rstan__ package:
+
+```{r, modelfit}
+# Prepare data
+data(roaches)
+roaches$roach1 <- sqrt(roaches$roach1)
```
paul-buerkner: what is the logic for using the square root of roach1?

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,309 @@
...
+y <- roaches$y
+x <- roaches[,2:4]
+offset <- log(roaches[,5])
+n <- dim(x)[1]
+k <- dim(x)[2]
+
+standata <- list(N = n, K = k, x = as.matrix(x), y = y,offset = offset,beta_prior_scale = 2.5,alpha_prior_scale = 5.0)
```
paul-buerkner: add some spaces after the `,` in function arguments

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,309 @@
...
+by the moment matching. These can be considered as diagnostics
+for how big influence each observation has on the posterior distribution.
+In addition to the Pareto $k$ diagnostics, moment matching also updates
+the effective sample size estimates. 
+
+# Using `loo_moment_match()` directly
+
+The moment matching can also be performed by explicitly calling the function
+`loo_moment_match()`. This enables its use also for models that
+are not using __rstan__.
```
paul-buerkner: or another package with built-in support for `loo_moment_match()`

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,309 @@
...
+To use `loo_moment_match()`, the user must give the model object `x`, the `loo`
+object, and 5 helper functions as arguments to
+`loo_moment_match()`.
+These helper functions, in turn, take `x` as an argument.
```
paul-buerkner: Since you explain the arguments of each function below, stating here that these helper functions take `x` as an argument is redundant and might be misleading since we have more arguments.

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,309 @@
...
+The helper functions are
+
+* `post_draws`	
+    + A function the takes `x` as the first argument and returns a matrix of posterior draws of the model parameters, `pars`.
+* `log_lik_i`	
+    + A function that takes `x` and `i` and returns a matrix (one column per chain) or a vector (all chains stacked) of log-likeliood draws of the ith observation based on the model x. If the draws are obtained using MCMC, the matrix with MCMC chains separated is preferred.
```
paul-buerkner: there is one x which should be quoted as `x`

vignettes/loo2-moment-matching.Rmd
```diff
@@ -0,0 +1,309 @@
...
+object.
+
+```{r, message=FALSE}
+loo3 <- loo::loo_moment_match.default(x = fit, loo = loo1,
+                              post_draws = post_draws_stanfit,
+                              log_lik_i = log_lik_i_stanfit,
+                              unconstrain_pars = unconstrain_pars_stanfit,
+                              log_prob_upars = log_prob_upars_stanfit,
+                              log_lik_i_upars = log_lik_i_upars_stanfit)
+loo3
```
paul-buerkner: Shall we add a sentence here that the results are, of course, the same as above?

vignettes/loo2-moment-matching.Rmd
topipa: I just made it similar to the case study https://avehtari.github.io/modelselection/roaches.html 

@avehtari Do you think this is better than `roaches$roach1 <- roaches$roach1 / 100` which is used in http://mc-stan.org/loo/articles/loo2-example.html ?

avehtari: Either is fine. sqrt seems to provide only slightly better result, so it doesn't matter which one is used.

jgabry:
Thanks this looks good! The only issue I see is that we need to wait to release this onto CRAN until `loo(fit, moment_match = TRUE)` is possible with the loo method in rstan, right?

topipa:
Yes, you're right, I didn't think of that. One solution could be to not run that part for now but simply print what the output would look like. I wouldn't want to fully remove it because I think it would be good to show that moment matching will be easy to use with rstan eventually without requiring these helper functions.

jgabry:
If you want to change it temporarily so that it's not evaluated and we just show what the output would look like that's ok by me as a temporary solution. Or we can wait for rstan.

avehtari:
> Or we can wait for rstan.

I think it's better no to wait, as there can be unexpected delays and we could get some feedback before that even if it requires copy pasting the code to define the functions needed.

topipa:
Okay, that part is now not evaluated so the vignette should work with current version of RStan.

PR #141, opened 2020-03-10
Parallelise moment matching

topipa:
This PR fixes parallelisation of moment matching. Now if `cores > 1`, moment matching is run in parallel for different problematic observations. Also the code is neater because the for loop is replaced with lapply.


jgabry:
Thanks, looks good to me too!

PR #146, opened 2020-06-10
return unshifted log weights in sis, tis, and psis

avehtari:
Instead of just documenting the old behavior, fix the behavior so that sis, tis, and psis can be used to estimate the normalization term as discussed in #112


MansMeg:
When reading the suggested changes I think it looks good. 

Although I think it would be very good to check the additional test (that now failed) and actually compute them more or less manually so we know that these changes actually give us the right result.  With this I mean to go through the examples that now fail and compute what we would expect the results to be. I added the following test suite  (that now fails). Hence these needs to be corrected with this new implementation:

```
  [ OK: 1009 | SKIPPED: 0 | WARNINGS: 0 | FAILED: 5 ]
  1. Failure: psis results haven't changed (@test_psis.R#17) 
  2. Failure: explict test of values for 'sis' and 'tis' (@test_tisis.R#129) 
  3. Failure: explict test of values for 'sis' and 'tis' (@test_tisis.R#132) 
  4. Failure: explict test of values for 'sis' and 'tis' (@test_tisis.R#138) 
  5. Failure: explict test of values for 'sis' and 'tis' (@test_tisis.R#142) 
```

The first fail is just that this change the reference example. The last four tests are the interesting ones:
https://github.com/stan-dev/loo/blob/master/tests/testthat/test_tisis.R

I get the feeling that we should probably have similar test for PSIS, but that might need a little more work. So maybe just assign that as an issue?

avehtari:
test values should differ only by the offset, which is the same as the largest value returned by the new code.

MansMeg:
Can you try to add the offset at line 130, 133, 139 and 143 in this file:
https://github.com/stan-dev/loo/blob/master/tests/testthat/test_tisis.R
At line 143 I think you can just remove the -max(lw) and push to this PR?

MansMeg:
Ok, now only a reference test fails:
```
  ── 1. Failure: psis results haven't changed (@test_psis.R#17)  ─────────────────
  `psis1` has changed from known value recorded in 'reference-results/psis.rds'.
  Attributes: < Component "norm_const_log": Mean relative difference: 0.3763288 >
  Component "log_weights": Mean relative difference: 1.403483
```

It is a reasonable difference and we should expect this change. Although, I'm not sure how to update the reference piss object? Maybe @jgabry knows?

jgabry:
Ok everything is passing now. Merging.

PR #147, opened 2020-06-10
fix #116 by making mcse_elpd_loo estimate deterministic

### LINKED ISSUE #116: loo results not reproducible between calls (loo$pointwise$mcse_elpd_loo field)
Two calls to `loo()` on the same object will not produce identical results, as `mcse_elpd()` calls `rnorm()` without setting a seed. Perhaps this is by design, but if so, it's not documented. This affects only the `mcse_elpd_loo` column of `loo$pointwise`.

avehtari:
fixes #116 
instead of rnorm, use approximate ordered statistics with offset recommended by Blom (1958)
Gunnar Blom. Statistical Estimates and Transformed Beta-Variables. Wiley; New York, 1958.


topipa:
Thanks for checking these tests! The difference was expected, because that test called rnorm without setting a random seed. Therefore, previous calls to loo had changed the state of the random number generator. Now when the call to rnorm was removed from loo, the state is different. So it should be nothing to worry about. I added a random seed to the failing test and recreated the reference result.

jgabry:
Great, thanks @topipa! 

Ok, so now the moment matching tests all pass, we just need to sort out the subsampling tests. I'm pretty sure that nothing is broken, but it seems like rounding here is making some things round up and others round down and then the difference is 1 instead of 0. 

@MansMeg does it make sense not to round and to compare the results for equality with some allowed tolerance instead? For example, instead of 

```r
expect_equal(round(lcss2[,1]), round(lcsso[,1]))
```

we could have something like:

```r
expect_equal(lcss2[,1], lcsso[,1], tolerance = 1)
```

MansMeg:
Hi!

Yes and yes. I think you can change the tests to use tolerance instead. I also think this might be such an error. Especially when the other tests indicate that it is quite a large difference for different seeds. 

I do not know how to push stuff into this PR? So if anyone just changes the tests to the new returned values it should probably work out fine.

jgabry:
@MansMeg I think you can just switch to the branch `fix-116-mcse_elpd-estimate` and pushing will push to the PR (that's what I did).

jgabry:
I'm happy to fix it too, but @MansMeg when you say

> So if anyone just changes the tests to the new returned values it should probably work out fine.

do you mean something other than switching to using a tolerance level?

MansMeg:
Nope. I just did that and it seemed to work. And now I realized how easy it was to push commits into a PR.  You learn something new every day. =)

jgabry:
Awesome, thanks! I'll let appeveyor finish running and then merge this

PR #149, opened 2020-07-08
Fix missing dots in 'loo_moment_match'

paul-buerkner:
This PR fixes a typo in `loo_moment_match` that prevented `...` to be passed to the `loo_moment_match_split` method. This causes trouble as soon as any additional argument is passed to the helper functions of moment matching. I should have seen this earlier (before the loo release) but unfortunately I didn't. 

@jgabry we should make another small CRAN update as soon as possible to make moment matching generally usable with additional arguments.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: loo
 Type: Package
 Title: Efficient Leave-One-Out Cross-Validation and WAIC for Bayesian Models
-Version: 2.3.0.9000
+Version: 2.3.0.9001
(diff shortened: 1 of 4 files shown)


jgabry:
Thanks for fixing. I'll have time to do another CRAN submission Thursday night or Friday. @topipa if you have time before then can you add a test for this on the `fix-missing-dots` branch?

topipa:
Test added. Thanks for the quick response on this!

jgabry:
Sorry for the delay on this. Resubmitting today.

PR #152, opened 2020-07-29
Fix problems with relative_eff.function

paul-buerkner:
This fixes the issue described in https://discourse.mc-stan.org/t/problems-running-loo-on-1-core/13908/22

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: loo
 Type: Package
 Title: Efficient Leave-One-Out Cross-Validation and WAIC for Bayesian Models
-Version: 2.3.1
+Version: 2.3.1.9000
 Date: 2020-07-14
 Authors@R: c(person("Aki", "Vehtari", email = "Aki.Vehtari@aalto.fi", role = c("aut")),
              person("Jonah", "Gabry", email = "jsg2201@columbia.edu", role = c("cre", "aut")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,18 @@
+# loo 2.3.1.9000
+
+<!-- Items for next release go here -->
+
+(GitHub issue/PR number in parentheses)
+
+* Fixed a bug in `relative_eff.function()` that caused an error on Windows when
+using multiple cores. (#152) 
+
+
 # loo 2.3.1
 
 (GitHub issue/PR number in parentheses)
 
-* Fix a bug in `loo_moment_match()` that prevented `...` arguments from being used
+* Fixed a bug in `loo_moment_match()` that prevented `...` arguments from being used
 correctly. (#149)
 
 
--- a/R/effective_sample_sizes.R
+++ b/R/effective_sample_sizes.R
@@ -124,6 +124,7 @@ relative_eff.function <-
           )
       } else {
         cl <- parallel::makePSOCKcluster(cores)
+        parallel::clusterExport(cl=cl, varlist=c("draws", "chain_id", "data"), envir=environment())
         on.exit(parallel::stopCluster(cl))
         n_eff_list <-
           parallel::parLapply(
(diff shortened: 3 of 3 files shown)


jgabry:
Thanks for helping with this. This seems to cause appveyor to fail with 

```
2 nodes produced errors; first error: object 'draws' not found
```

The reason I added `parallel::clusterExport(cl, "draws")` in the first place was because that seemed to avoid an appveyor error, but adding it seems to cause errors for users (like reported on discourse). So I'm confused ;)

paul-buerkner:
`parallel::clusterExport(cl, "draws")` looks in  the global env for "draws" by default which is not what you want. You could change `envir = environment()` in the call.

rok-cesnovar:
Should the call also export data and chain_id ?

clusterExport(cl=cl, varlist=c("draws", "chain_id", "data"), envir=environment())

paul-buerkner:
If there is a problem with finding these objects then yes, but so far I
don't think this is a problem on regular windows machines just on the
testing engine apparently

Rok Češnovar <notifications@github.com> schrieb am Mi., 29. Juli 2020,
20:17:

jgabry:
It would be nice to have a solution that works both on users machines and also in the tests. In that case, should we try 

```r
clusterExport(cl=cl, varlist=c("draws", "chain_id", "data"), envir=environment())
```

and see if that works both on users machines _and_ appveyor?

paul-buerkner:
Sounds good to me. Would you mind making this change yourself as I will not
be at my laptop tonight?

Am Mi., 29. Juli 2020 um 20:40 Uhr schrieb Jonah Gabry <
notifications@github.com>:

jgabry:
Yeah no problem I can do it

jgabry:
Ok pushed. Let's see if appveyor passes now and if it does we can check if it also works locally for individual windows users.

rok-cesnovar:
Do you have a minimal example to test this? I can run it on my Windows machine.

paul-buerkner:
Any rstanarm model with loo(model, cores = 2) should do. I am also happy to
change and test tomorrow if you are willing to wait until then.

Rok Češnovar <notifications@github.com> schrieb am Mi., 29. Juli 2020,
20:58:

jgabry:
@rok-cesnovar Thanks, can you try running this example from the loo doc? 

```r
# Simulate data and draw from posterior
N <- 50; K <- 10; S <- 100; a0 <- 3; b0 <- 2
p <- rbeta(1, a0, b0)
y <- rbinom(N, size = K, prob = p)
a <- a0 + sum(y); b <- b0 + N * K - sum(y)
fake_posterior <- as.matrix(rbeta(S, a, b))
dim(fake_posterior) # S x 1
fake_data <- data.frame(y,K)
dim(fake_data) # N x 2

llfun <- function(data_i, draws) {
  # each time called internally within loo the arguments will be equal to:
  # data_i: ith row of fake_data (fake_data[i,, drop=FALSE])
  # draws: entire fake_posterior matrix
  dbinom(data_i$y, size = data_i$K, prob = draws, log = TRUE)
}

reff <- loo::relative_eff(llfun, chain_id = rep(1, S), cores = 2, 
                          data = fake_data, draws = fake_posterior)
loo_with_fn <- loo::loo.function(llfun, r_eff = reff, cores = 2, 
                                 draws = fake_posterior, data = fake_data)
```

jgabry:
Or like @paul-buerkner loo with cores>1 after any rstanarm model

jgabry:
Looks like it passes on appveyor now

rok-cesnovar:
On master I get the `object 'draws' not found` error. On this branch everything works.

jgabry:
Great, thanks for checking!

jgabry:
I think we can go ahead and merge this now. At a minimum it won't make anything worse, and it's working both for @rok-cesnovar locally and on appveyor.

jgabry:
I'll update the NEWS file and then merge this

PR #153, opened 2020-09-17
Fixed a numerical issue in split_moment_matching

topipa:
This PR fixes the issue https://github.com/stan-dev/rstan/issues/852 .

When computing the density of the split proposal, we need to compute a
logarithm of the sum of two densities `log(a + c)`. In some extreme cases, to compute this in a stable way we need a different formula depending on whether `a >> c` or `a << c`.

### DIFF
--- a/R/split_moment_matching.R
+++ b/R/split_moment_matching.R
@@ -82,26 +82,21 @@ loo_moment_match_split <- function(x, upars, cov, total_shift, total_scaling,
   log_liki_half <- log_lik_i_upars(x, upars = upars_trans_half, i = i, ...)
 
   # compute weights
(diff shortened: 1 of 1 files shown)


MansMeg:
I guess the easiest is if @ave or @paulbuerkner review this PR?

jgabry:
Thanks Topi! 

> I guess the easiest is if @avehtari or @paul-buerkner review this PR?

That makes sense. Aki and Paul if you don't have time let us know.

paul-buerkner (approved):
Looks good to me.

jgabry:
Thanks @topipa and @paul-buerkner!

Issue #151, opened 2020-07-16
loo_moment_match integer overflow


trinhdhk:
*Problem*: Integer Overflow when using loo_moment_match()

**What I did:**
- *loo_moment_match* for a stanfit model, following the vignette [here](https://mc-stan.org/loo/articles/loo2-moment-matching.html), with all default parameters and functions.
- Number of cores for parallelisation: 10

**Expected behaviour:** return the value with no exceptions

**Actual behaviour:**  return an error message
```
Error in while (t < nrow(acov) - 5 && !is.nan(rho_hat_even + rho_hat_odd) &&  : 
  missing value where TRUE/FALSE needed
In addition: Warning message:
In N * N : NAs produced by integer overflow
```

**Is the problem random?** No

**What I tried:** 
- Reduce the number of cores *didn't work*
- Remove the high-dimensional random effect parameter *didn't work, the parameter is required* 

**Session Info**
```
R version 4.0.2 (2020-06-22)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 18.04.4 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.7.1
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.7.1

locale:
 [1] LC_CTYPE=C.UTF-8       LC_NUMERIC=C           LC_TIME=C.UTF-8        LC_COLLATE=C.UTF-8    
 [5] LC_MONETARY=C.UTF-8    LC_MESSAGES=C.UTF-8    LC_PAPER=C.UTF-8       LC_NAME=C             
 [9] LC_ADDRESS=C           LC_TELEPHONE=C         LC_MEASUREMENT=C.UTF-8 LC_IDENTIFICATION=C   

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] bayesplot_1.7.2      rstan_2.19.3         ggplot2_3.3.2        StanHeaders_2.21.0-5
[5] magrittr_1.5         data.table_1.12.8   

loaded via a namespace (and not attached):
 [1] tinytex_0.24       tidyselect_1.1.0   xfun_0.15          purrr_0.3.4        colorspace_1.4-1  
 [6] vctrs_0.3.1        generics_0.0.2     stats4_4.0.2       loo_2.3.1          rlang_0.4.7       
[...]

PR #155, opened 2020-09-22
fix integer overflow from #151

ecmerkle:
I encountered this problem myself and then found #151, so added the trivial fix that @avehtari describes. (but I realize the posterior package might make it irrelevant)


jgabry (approved):
Thanks @ecmerkle! This looks good to me. @avehtari any reason we shouldn't merge this? We can still switch to calling the functions from `posterior` when on CRAN, but good to have this until then.

avehtari:
It's good to merge this!

PR #161, opened 2020-11-26
Fix relative_eff with draws_array

rok-cesnovar:
This fixes https://github.com/stan-dev/posterior/issues/101 though there might be cleaner solutions. Feel free to close and push a cleaner solution if there is one.

### DIFF
--- a/R/effective_sample_sizes.R
+++ b/R/effective_sample_sizes.R
@@ -70,7 +70,7 @@ relative_eff.array <- function(x, ..., cores = getOption("mc.cores", 1)) {
         parallel::mclapply(
           mc.cores = cores,
           X = seq_len(dim(x)[3]),
-          FUN = function(i) ess_rfun(x[, , i])
+          FUN = function(i) ess_rfun(x[, , i, drop = TRUE])
         )
     } else {
       cl <- parallel::makePSOCKcluster(cores)
@@ -79,7 +79,7 @@ relative_eff.array <- function(x, ..., cores = getOption("mc.cores", 1)) {
         parallel::parLapply(
           cl = cl,
           X = seq_len(dim(x)[3]),
-          fun = function(i) ess_rfun(x[, , i])
+          fun = function(i) ess_rfun(x[, , i, drop = TRUE])
         )
     }
     n_eff_vec <- unlist(n_eff_list, use.names = FALSE)
@@ -209,11 +209,7 @@ ess_rfun <- function(sims) {
   if (is.vector(sims)) dim(sims) <- c(length(sims), 1)
   chains <- ncol(sims)
   n_samples <- nrow(sims)
-  if (length(dim(sims)) == 3) {
-    acov <- lapply(1:chains, FUN = function(i) autocovariance(sims[,i,]))
-  } else {
-   acov <- lapply(1:chains, FUN = function(i) autocovariance(sims[,i]))
-  }
+  acov <- lapply(1:chains, FUN = function(i) autocovariance(sims[,i]))
   acov <- do.call(cbind, acov)
   chain_mean <- colMeans(sims)
   mean_var <- mean(acov[1,]) * n_samples / (n_samples - 1)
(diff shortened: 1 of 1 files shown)


rok-cesnovar:
Tested with the following:

```
library(posterior)
library(loo)
options(mc.cores = 2)
mu <- posterior::subset_draws(posterior::example_draws("eight_schools"), variable = c("mu"))
loo::relative_eff(mu)

options(mc.cores = 1)
mu <- posterior::subset_draws(posterior::example_draws("eight_schools"), variable = c("mu"))
loo::relative_eff(mu)

options(mc.cores = 1)
mu <- posterior::subset_draws(posterior::example_draws("eight_schools"), variable = c("mu"))
class(mu) <- "array"
loo::relative_eff(mu)

options(mc.cores = 2)
mu <- posterior::subset_draws(posterior::example_draws("eight_schools"), variable = c("mu"))
class(mu) <- "array"
loo::relative_eff(mu)

options(mc.cores = 2)
mu <- posterior::subset_draws(posterior::example_draws("eight_schools"), variable = c("mu", "tau"))
loo::relative_eff(mu)

options(mc.cores = 1)
mu <- posterior::subset_draws(posterior::example_draws("eight_schools"), variable = c("mu", "tau"))
loo::relative_eff(mu)

options(mc.cores = 1)
mu <- posterior::subset_draws(posterior::example_draws("eight_schools"), variable = c("mu", "tau"))
class(mu) <- "array"
loo::relative_eff(mu)

options(mc.cores = 2)
mu <- posterior::subset_draws(posterior::example_draws("eight_schools"), variable = c("mu", "tau"))
class(mu) <- "array"
loo::relative_eff(mu)
```
Given that posterior is not on CRAN I could not add any tests for this case.

jgabry:
Thanks @rok-cesnovar! This solution definitely works and it might be the best option for now. 

I also think we should think a bit more about how to handle errors like this because the way posterior works could result in more situations like this in any package that ends up supporting both regular array objects and draws_array objects.  

> there might be cleaner solutions

I did have a cleaner solution in mind but my idea doesn't seem to work with the way posterior is designed. What I was thinking was that we shouldn't pass a 3-D array to ess_rfun (which expects a 2-D array), we should instead just drop the extra dimension first, but if I change this

https://github.com/stan-dev/loo/blob/385a1ba2fb7d22c75b795866ff167608a13d2b9f/R/effective_sample_sizes.R#L79-L83

to 

```r
fun = function(i) ess_rfun(x[, , i, drop = TRUE])
```

then posterior just ignores the drop=TRUE and we get the same error. @paul-buerkner What `ess_rfun` really wants as input is the output from `posterior::extract_variable_matrix()`, but we can't yet call that function inside of loo because posterior isn't on CRAN, and even if we could call it that function requires explicitly naming the variable to extract but I don't think loo should worry about variable names. 

While drop=FALSE is a good default for posterior, if it doesn't also respect drop=TRUE then I think other packages will have to use hacky ways to support both regular arrays and posterior draws_arrays (e.g. branching on the length of dim when they wouldn't otherwise have to do that), right? But maybe there would be other problems if we allowed drop=TRUE to work in cases like this. I haven't thought it through too much (we also probably had a discussion like this over in the posterior repo at some point). 

I think ultimately for loo and our other R packages the solution could be to replace all the internals to expect objects from posterior (if the user passes in a regular array then we immediately convert it to a draws_array and all the internal functions are rewritten expect the behavior of draws_arrays instead of branching to allow for both types of arrays). But that doesn't really help other package authors who will want to support both regular arrays and draws_array objects. 

Does anyone have any thoughts on this or any other ideas for getting around this problem? In many ways this is a super minor problem but it actually raises some issues that might affect us and other package authors in the future.

paul-buerkner:
I think that we should allow drop = TRUE but if dimensions are actually
dropped, the result will not be labeled as class `draws_array`anymore but
simply of whatever R would usually do in that case. I did the same already
with draws_df objects when meta columns were dropped, such that the format
is no longer valid (instead it will just come out as a tibble). Does that
make sense to you?

Am Do., 26. Nov. 2020 um 19:29 Uhr schrieb Jonah Gabry <
notifications@github.com>:

jgabry:
Ok great, yeah that makes sense.

jgabry:
Ok I just changed this PR to use a slightly different solution where we just drop the extra dimension instead of branching based on the number of dimensions. This doesn't work with the master branch of posterior but does work now with https://github.com/stan-dev/posterior/pull/102.

jgabry:
I think we can merge this now that the corresponding PR in posterior was merged.

PR #176, opened 2021-05-22
Clarified documentation

MansMeg:
Updated the documentation on the subsampling loo to clarify the double use of the log-likelihood function.

### DIFF
--- a/R/loo_subsample.R
+++ b/R/loo_subsample.R
@@ -34,7 +34,9 @@ loo_subsample <- function(x, ...) {
 #' @template function
 #' @param data,draws,... For `loo_subsample.function()`, these are the data,
 #'   posterior draws, and other arguments to pass to the log-likelihood
-#'   function.
+#'   function. Note that for some `loo_approximation`s, the draws will be replaced
+#'   by the posteriors summary statistics to compute loo approximations. See
+#'   `loo_approximation` for details.
 #' @param observations The subsample observations to use. The argument can take
 #'   four (4) types of arguments:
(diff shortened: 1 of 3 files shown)


R/loo_subsample.R
```diff
@@ -34,7 +34,9 @@ loo_subsample <- function(x, ...) {
 #' @template function
 #' @param data,draws,... For `loo_subsample.function()`, these are the data,
 #'   posterior draws, and other arguments to pass to the log-likelihood
-#'   function.
+#'   function. Note that for some `loo_approximation`s, the draws will be replaced
+#'   by the posteriors summary statistics to compute loo approximations. See
+#'   `loo_approximation` for details.
```
jgabry, line 39: Did you mean `loo_approximate_posterior` instead of `loo_approximation`? If so you can use `[loo_approximate_posterior()]` in the doc and it will auto link to that doc page.

MansMeg: Actually not. I was refering to the argument loo_approximation in the same function. Can that be clarified more somhow?

jgabry: Oh I just forgot that was the argument name. How about "See argument `loo_approximation` for details."?

MansMeg: Now fixed.

MansMeg:
@kdubovikov would this fix the issue you had with the docs?

kdubovikov:
Yes, this is clear now

jgabry (approved):

Issue #127, opened 2019-10-22
plot method for two loo objects


jgabry:
We should have a plotting method (in loo or bayesplot) that creates the pointwise elpd difference plot from Figure 10a of [Visualization in Bayesian Workflow](https://arxiv.org/abs/1709.01449)

jgabry:
Also have an argument to allow reordering the data points in some meaningful way (defaulting to current order)

PR #178, opened 2021-05-28
Add LOO Difference Plot

ParadaCarleton:
Implement plot from Bayesian workflow paper


ParadaCarleton:
Change file name

ParadaCarleton:
Closes #127

R/loo_difference_plot.R
```diff
@@ -0,0 +1,151 @@
...
+           psis_object_1,
+           psis_object_2,
+           ...,
+           group = FALSE,
+           outlier_thresh = FALSE,
+           size = 1,
+           alpha = 1,
+           jitter = 0,
+           quantiles = FALSE,
+           sortByGroup = FALSE
```
jgabry, line 75: I would use `sort_by_group` instead of `SortByGroup` to match the style already used in the package (and also in this function, e.g. `psis_object_1`)

R/loo_difference_plot.R
```diff
@@ -0,0 +1,151 @@
...
+#' @examples 
+#' 
+#' library(loo)
+#' 
+#' cbPalette <- c("#636363", "#E69F00", "#56B4E9", "#009E73", 
+#'                "#F0E442", "#0072B2","#CC79A7")
+#' 
+#' # Plot using groups from WHO
+#' 
+#' plot_loo_dif(factor(GM@data$super_region_name), loo3, loo2,
```
jgabry, line 42: It looks like here in the doc it's `plot_loo_dif` but in the code it's `plot_loo_variation`.

jgabry, line 42: Right now none of these examples run because the data `GM` and loo objects (`loo2`, `loo3`) don't exist. For the Examples section in the doc we'll need to change this to a self contained example or add all this data to the loo package itself so it can be used in the example. One possibility is to use a toy example here in the doc and then potentially add a more real example (e.g. this one from the paper) in one of the package vignettes. I'll think a bit more about this too.

R/loo_difference_plot.R
```diff
@@ -0,0 +1,151 @@
...
+#'              xlab("Index") + scale_colour_manual(values=cbPalette)
+#' 
+#' 
+plot_loo_variation <- 
+  function(y,
+           psis_object_1,
+           psis_object_2,
+           ...,
+           group = FALSE,
+           outlier_thresh = FALSE,
```
jgabry, line 70: I don't think this argument has documentation

R/loo_difference_plot.R
```diff
@@ -0,0 +1,151 @@
...
+             ")
+      }
+      
+      values <- group_by(tibble(group, elpdDif), factor(group)) %>%
+                    arrange(.by_group = TRUE)
+      
+      elpdDif <- pull(values, elpdDif)
+      group <- pull(values, group)
+      
+    }
```
jgabry, line 106: It looks like this is using tibble and dplyr code but we should avoid that if possible (I like those packages in general but for package development we don't want to add two new dependencies to the loo package for functionality that can be accomplished without those dependencies, even if the code is a bit more cumbersome).

R/loo_difference_plot.R
```diff
@@ -0,0 +1,151 @@
...
+    
+    
+    plot <- ggplot(mapping=aes(y, elpdDif)) +
+            geom_hline(yintercept=0) + 
+            xlab(ifelse(sortByGroup, "y", "Index")) +
+            ylab(expression(ELPD[i][1] - ELPD[i][2])) + 
+            labs(color = "Groups")
```
jgabry, line 113: In order to call functions from ggplot (or any other package) inside of the loo package source code we need to prefix them with `ggplot2::`, e.g. `ggplot2::geom_hline()`.

R/loo_difference_plot.R
```diff
@@ -0,0 +1,151 @@
...
+#'              ) + 
+#'              xlab("Index") + scale_colour_manual(values=cbPalette)
+#' 
+#' 
+plot_loo_variation <- 
+  function(y,
+           psis_object_1,
+           psis_object_2,
+           ...,
+           group = FALSE,
```
jgabry, line 69: I would use `NULL` instead of `FALSE`. To me a default of `FALSE` gives the impression that the possible values are TRUE and FALSE, but `group=TRUE` isn't a possible value. So I think `NULL` makes more sense as the default. Same with `outlier_thresh`.   On the other hand, `FALSE` makes sense for `quantiles` since it seems to actually be an argument that needs a logical/boolean value.

R/loo_difference_plot.R
jgabry: For example, here are some calls to the parallel package from a different file in the loo package: 

https://github.com/stan-dev/loo/blob/01178587e101c4e06e5ab078acdf69ab272a75f2/R/loo_moment_matching.R#L130-L133

We need to do the same thing with ggplot2.

jgabry (dismissed):
Hey Carlos, thanks for working on this! There are a bunch of small things I put in review comments (most of it related to things that matter for package development but not necessarily when writing code outside of package development), but this is a really really great start. Once we sort out some of smaller issues I commented on we can then get some other people, e.g. @avehtari or the other loo package authors to also take a look and see what they think about the plot itself and the different options provided.

R/loo_difference_plot.R
ParadaCarleton: I believe this is the only problem that's left unresolved; as I mentioned in the email, I'd like to see if it's possible to add this data to the LOO package, since I think the example here is really great.

R/loo_difference_plot.R
```diff
@@ -0,0 +1,151 @@
...
+#' to the corresponding value of `y`. If `FALSE`, ignored.
+#' @param outlier_thresh Flag values when the difference in the ELPD exceeds
+#' this threshold. Defaults to `NULL`, in which case no values are flagged.
+#' @param size,alpha,jitter Passed to `[ggplot2::geom_point()]` to control 
+#' aesthetics. `size` and `alpha` are passed to to the `size` and `alpha` 
+#' arguments of `[ggplot2::geom_jitter()]` to control the appearance of 
+#' points. `jitter` can be either a number or a vector of numbers.
+#' Passing a single number will jitter variables along the x axis only, while 
+#' passing a vector will jitter along both axes.
+#' @param quantiles Boolean that determines whether to plot the quantiles of
```
avehtari, line 23: For me `quantiles` sounds like it would accept vector of quantiles (real valued) as, e.g., dnorm and pnorm. I suggest that `quantiles` parameter would actually allow to define which quantiles are plotted and the type of the plot would be selected with argument called, e.g., `type`. With options type='y' (default) and type='quantiles', which would allow then also extending to other possible types- Can you also include example of how these plots look like in the discussion thread of this PR?

ParadaCarleton: > For me `quantiles` sounds like it would accept vector of quantiles (real valued) as, e.g., dnorm and pnorm. 
This does seem like it might be confusing; I think it makes sense to remove this, since users can provide transformations of the values themselves by replacing `y`.

> I suggest that `quantiles` parameter would actually allow to define which quantiles are plotted and the type of the plot would be selected with argument called, e.g., `type`. With options type='y' (default) and type='quantiles', which would allow then also extending to other possible types
I'm not sure what you mean here; could you elaborate?

> Can you also include example of how these plots look like in the discussion thread of this PR?
Will do.

ParadaCarleton:
@avehtari I've added an example of a plot using the IQ dataset from rstanarm:
image
The first model only has a constant intercept; the second model also includes a term for the mother's IQ. At the edges, the constant-based model is more accurate than the full model, while the IQ-based model is stronger in the center. This seems counterintuitive, until you realize that this reveals there is significant measurement error in IQ tests: extreme values are probably caused by mismeasurement and need to be adjusted for regression to the mean.

Other examples:
unnamed

R/loo_difference_plot.R
avehtari: If users can provide transformations of the values themselves by replacing y, and this option is not needed then I think removing it as you did is the correct action and you don't need to care about the rest I said.

avehtari:
The plots looks great! Thanks for making this PR!

ParadaCarleton:
> The plots looks great! Thanks for making this PR!

Thanks! Do you happen to know if the spatial data from the Bayesian visualization paper can be added to the loo package? I think it's a good example, but if it can't, I can remove it.

jgabry:
Sorry for the delay in responding to this. Regarding the issue of including the data from the visualization paper, I think we should avoid adding it to the package but I think there's still a way we can still use it. Basically, we should use a different example in the Examples section in the documentation, but then we can add an example using the visualization paper data in one of the vignettes. This works because in the vignette we're allowed to download data that isn't included in the package. For example, we do this already in one of the vignettes with some other data: 

https://github.com/stan-dev/loo/blob/01178587e101c4e06e5ab078acdf69ab272a75f2/vignettes/loo2-with-rstan.Rmd#L95-L96

So we could do something similar and download the visualization paper data from https://github.com/jgabry/bayes-vis-paper. @ParadaCarleton What do you think about that option?

ParadaCarleton:
> Sorry for the delay in responding to this. Regarding the issue of including the data from the visualization paper, I think we should avoid adding it to the package but I think there's still a way we can still use it. Basically, we should use a different example in the Examples section in the documentation, but then we can add an example using the visualization paper data in one of the vignettes. This works because in the vignette we're allowed to download data that isn't included in the package. For example, we do this already in one of the vignettes with some other data:
> 
> https://github.com/stan-dev/loo/blob/01178587e101c4e06e5ab078acdf69ab272a75f2/vignettes/loo2-with-rstan.Rmd#L95-L96
> 
> So we could do something similar and download the visualization paper data from https://github.com/jgabry/bayes-vis-paper. @ParadaCarleton What do you think about that option?

Sounds good! I've removed the example. Anything else?

jgabry:
> Where do you think I should put this -- should I add it to an existing vignette or create a new one?

Hmm, on the one hand I think it could be nice to have it in the main introduction vignette https://mc-stan.org/loo/articles/loo2-example.html. On the other hand, this would be using a totally different data set than the example in that vignette, so that's not ideal. A new vignette would be great, but I think if we go that route then the new vignette shouldn't just be about this one plot but rather about visualizing loo output in general (e.g., this plot, the Pareto k diagnostic plot, perhaps some of the loo related plots in bayesplot, etc.). But that would be more work than just adding this to an existing vignette so it depends how much you feel like working on this (no pressure!).

ParadaCarleton:
@jgabry btw, I think we should merge this, unless there's some change you'd like to see. I might be able to get to building a vignette before classes start again, but given that I'm focused on adding new features to ParetoSmooth.jl I'm not sure I'll be able to. We can probably add this to some other vignette later.

avehtari:
@ParadaCarleton and @jgabry , what is the status of this PR?

VisruthSK:
My changes were assisted by ChatGPT (almost exclusively for writing code I didn't want to type out, not for ideas):

Renamed args to `loo_1` and `loo_2`, and other small renames.
Removed vertical jitter and made labels jitter with points so they would be aligned.
Changed labelling args/semantics, don't call it outlier thresholding.
Added some arg checking.
Added tests.

R/loo_difference_plot.R
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    if (!is.null(labels)) {
+      checkmate::assert_atomic_vector(
+        labels,
+        len = length(y)
+      )
+    }
+
+    if (!is.null(label_threshold) && is.null(labels)) {
```
jgabry, line 129: @VisruthSK If `labels` is provided but `label_threshold` isn't then labels are just ignored I think. I'm thinking we should maybe error in that case?

VisruthSK: > `label_threshold` must be supplied when `labels` is supplied.

Do you think this is clear enough?

VisruthSK:
Note: the nice plots from the example should probably be added back in a vignette. Will work on that after the actual plotting function is approved.

R/loo_difference_plot.R
jgabry: Yeah I think so

jgabry:
> Note: the nice plots from the example should probably be added back in a vignette. Will work on that after the actual plotting function is approved.

Sounds good. I saw you converted back to a draft. Should I wait to take another look?

VisruthSK:
Fine to take a look, but will leave as draft until vignette is in--or cut vignette, or make vignette its own PR.

VisruthSK:
Also note: right now there are no visual snaps since that would add vdiffr as a dep.

PR #187, opened 2021-09-20
simplify the normalization in gpdfit

avehtari:
Seth Axen pointed out that 
```
w_theta <- 1 / vapply(jj, FUN.VALUE = numeric(1), FUN = function(j) {
  sum(exp(l_theta - l_theta[j]))
})
```
Is doing normalization in O(n^2), while it can be simplified to O(n) and also made more stable as
```
w_theta <- exp(l_theta - matrixStats::logSumExp(l_theta))
```

### DIFF
--- a/R/gpdfit.R
+++ b/R/gpdfit.R
@@ -40,9 +40,7 @@ gpdfit <- function(x, wip = TRUE, min_grid_pts = 30, sort_x = TRUE) {
   xstar <- x[floor(N / 4 + 0.5)] # first quartile of sample
   theta <- 1 / x[N] + (1 - sqrt(M / (jj - 0.5))) / prior / xstar
   l_theta <- N * lx(theta, x) # profile log-lik
-  w_theta <- 1 / vapply(jj, FUN.VALUE = numeric(1), FUN = function(j) {
-    sum(exp(l_theta - l_theta[j]))
-  })
+  w_theta <- exp(l_theta - matrixStats::logSumExp(l_theta)) # normalize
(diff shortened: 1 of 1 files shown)


ParadaCarleton:
@sethaxen Is this taken from some of the changes I made while writing ParetoSmooth.jl, or is it a separate performance improvement I should add to ParetoSmooth myself?

sethaxen:
@ParadaCarleton no, this isn't from ParetoSmooth.jl. If the improvement isn't in there, it should be.

jgabry:
This looks good, thanks! I need to investigate why the Windows checks are failing but haven't had a chance yet. I'm guessing it's unrelated to this PR, but want to double check before merging.

jgabry:
The windows test failure is happening here: 

https://github.com/stan-dev/loo/blob/bffab8cd46270cba8e7dd1d09c12b880dc8dda2c/tests/testthat/test_loo_and_waic.R#L123-L127

the error message in the [github actions log](https://github.com/stan-dev/loo/pull/187/checks?check_run_id=3666538192#step:11:174) says that the error occurs because the objects are "equal but not identical". This usually means that the value is the same but some attribute differs, but I can't think of any reason how this PR could cause that.  

Does anyone have access to a windows machine? If so could you try running [this one test](https://github.com/stan-dev/loo/blob/bffab8cd46270cba8e7dd1d09c12b880dc8dda2c/tests/testthat/test_loo_and_waic.R#L123-L127
) locally?

PR #190, opened 2021-11-16
Update loo2-with-rstan.Rmd

MansMeg:
Quick fix of arxiv link.


avehtari:
Thanks, that was fast!

PR #194, opened 2022-01-19
Minor fixes of references

MansMeg:
Hi!

I (think) I fixed the references to be correct. Since the (2017b) changed to (2019) we should also change the (2017a) ref to (2017). I Also updated the AISTAT ref when I was on it.


avehtari (approved):
Thanks!

PR #195, opened 2022-03-02
fix for CRAN and update NEWS

jgabry:
This PR prepares version 2.5.0 for release. CRAN requested a minor fix that requires a release. I think we should do this release (which includes all PRs merged since the last release) and then we can work on getting the remaining open PRs merged (sorry for the delay on those!). 

@avehtari Any further thoughts on the LOO-CV FAQ? Should we include that with the package in this release? Alternatively we can just add it to the LOO website without including it with the package source code (in which case we can do that at any time and we don't need to coordinate with a release).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,8 +1,8 @@
 Package: loo
 Type: Package
 Title: Efficient Leave-One-Out Cross-Validation and WAIC for Bayesian Models
-Version: 2.4.1.9000
-Date: 2020-12-07
+Version: 2.5.0
+Date: 2022-03-02
 Authors@R: c(person("Aki", "Vehtari", email = "Aki.Vehtari@aalto.fi", role = c("aut")),
              person("Jonah", "Gabry", email = "jsg2201@columbia.edu", role = c("cre", "aut")),
              person("Mans", "Magnusson", role = c("aut")),
@@ -51,5 +51,5 @@ Suggests:
 VignetteBuilder: knitr
 Encoding: UTF-8
 SystemRequirements: pandoc (>= 1.12.3), pandoc-citeproc
-RoxygenNote: 7.1.1
+RoxygenNote: 7.1.2
 Roxygen: list(markdown = TRUE)
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,20 +2,22 @@
 
 ### Improvements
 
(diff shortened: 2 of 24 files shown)


avehtari:
Is there some place in the package that if the CV-FAQ would be mentioned with a link to the web page, that would help finding the FAQ? Then mentioning it with link could be good, but have the content only on the web page?

Also we should update references in loo2-non-factorized, loo2-lfo, and loo2-moment-matching vignettes (I'll make an issue)

jgabry:
> Is there some place in the package that if the CV-FAQ would be mentioned with a link to the web page, that would help finding the FAQ? Then mentioning it with link could be good, but have the content only on the web page?
> 

I guess we could put a link in the glossary and in the See Also section of the doc for the `loo()` and `psis()` functions? If we're going to put a link in the package then we need the link to be active when we submit to CRAN (they actually automatically check for dead links), so we'd need to add the FAQ to the website before releasing. Or we could do this release, then add it to the website, and then add a link to it in the subsequent release. Do you have a preference? 

> Also we should update references in loo2-non-factorized, loo2-lfo, and loo2-moment-matching vignettes (I'll make an issue)

Ok great thank!

avehtari:
> so we'd need to add the FAQ to the website before releasing

I'm ok with this, but if this would cause too much work or we don't want to make mistakes in a hurry before you need to get the release to keep cran happy, then we can wait

jgabry:
I can try to get it on the website before this release and if we run into any issues we can wait (should be doable though). Do you have an Rmd file that has the source for the contents at https://avehtari.github.io/modelselection/CV-FAQ.html?

avehtari:
https://github.com/avehtari/modelselection/blob/master/CV-FAQ.Rmd

jgabry:
Ok thanks. I'll need to copy that file into the loo package so then I guess there will be two versions. Are you ok if the version that we maintain going forward is the one in the loo package and not the one in your repo? I guess we can also try to remember to always make the same updates in both, but that's not ideal.

jgabry:
@avehtari the FAQ is now live! 

https://mc-stan.org/loo/articles/online-only/faq.html

I'll add some links to it from the loo doc as part of this PR before I merge this and do the loo release

jgabry:
@avehtari I added links to the FAQ from the glossary and several other pages in the package doc. We can of course keep adding more links in the future wherever we see fit.

PR #202, opened 2022-05-19
Additional loo utilities

LeeviLindgren:
Hi!

This PR adds functions to compute additional LOO utilities: mean absolute error, mean squared error, and root mean squared error for regression problems and accuracy and balanced accuracy for binary classification. I have already discussed this with @avehtari and opened a PR for more discussion and feedback.


avehtari:
I think this PR could be merged adding just the loo_predictive_error function, but it would be useful to think the bigger picture and how we could support loo_compare with different criteria

jgabry:
Looking at this now. Sorry for the delay!

jgabry:
@LeeviLindgren Thanks for the PR! Sorry for the delay in reviewing it. I'll make a few review comments now, but can you also edits from maintainers? You can do that by following the instructions here:  https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

R/loo_predictive_error.R
```diff
@@ -0,0 +1,200 @@
...
+#' @examples
+#' \donttest{
+#' if (requireNamespace("rstanarm", quietly = TRUE)) {
+#' # Use rstanarm package to quickly fit a model and get both a log-likelihood
+#' # matrix and draws from the posterior predictive distribution
+#' library("rstanarm")
+#'
+#' # data from help("lm")
+#' d <- data.frame(
+#'   weight = c(ctl, trt),
```
jgabry, line 61: This currently results in an error because `ctl` and `trt` are not defined.

R/loo_predictive_error.R
```diff
@@ -0,0 +1,200 @@
...
+#'   group = gl(2, 10, 20, labels = c("Ctl","Trt"))
+#' )
+#' fit <- stan_glm(weight ~ group, data = d, refresh = 0)
+#' ll <- log_lik(fit)
+#' r_eff <- relative_eff(exp(-log_ratios), chain_id = rep(1:4, each = 1000))
```
jgabry, line 66: `log_ratios` isn't defined anywhere so this will error

R/loo_predictive_error.R
```diff
@@ -0,0 +1,200 @@
...
+      identical(ncol(x), length(y)),
+      identical(dim(x), dim(ll))
+    )
+    pred_error <- match.arg(pred_error)
+    psis_object <- psis(-ll, r_eff = r_eff)
+    pred_loo <- E_loo(x, psis_object = psis_object, log_ratios = -ll, ...)$value
+
+    predictive_error_fun <- .loo_predictive_error_fun(pred_error)
+
+    return(predictive_error_fun(y, pred_loo))
```
jgabry, line 103:
```suggestion
    predictive_error_fun(y, pred_loo)
```

R/loo_predictive_error.R
```diff
@@ -0,0 +1,200 @@
...
+#' Mean absolute error
+#'
+#' @noRd
+#' @param y A vector of observed values
+#' @param yhat A vector of predictions
+.mae <-function(y, yhat) {
+  stopifnot(length(y) == length(yhat))
+  n <- length(y)
+  e <- abs(y - yhat)
+  return(list(estimate = mean(e), se = sd(e) / sqrt(n)))
```
jgabry, line 135:
```suggestion
  list(estimate = mean(e), se = sd(e) / sqrt(n))
```

R/loo_predictive_error.R
```diff
@@ -0,0 +1,200 @@
...
+#' Mean squared error
+#'
+#' @noRd
+#' @param y A vector of observed values
+#' @param yhat A vector of predictions
+.mse <-function(y, yhat) {
+  stopifnot(length(y) == length(yhat))
+  n <- length(y)
+  e <- (y - yhat)^2
+  return(list(estimate = mean(e), se = sd(e) / sqrt(n)))
```
jgabry, line 147:
```suggestion
  list(estimate = mean(e), se = sd(e) / sqrt(n))
```

R/loo_predictive_error.R
```diff
@@ -0,0 +1,200 @@
...
+#' @param yhat A vector of predictions
+.accuracy <- function(y, yhat) {
+  stopifnot(length(y) == length(yhat),
+            all(y <= 1 & y >= 0),
+            all(yhat <= 1 & yhat >= 0))
+  n <- length(y)
+  yhat <- as.integer(yhat > 0.5)
+  acc <- as.integer(yhat == y)
+  est <- mean(acc)
+  return(list(estimate = est, se = sqrt(est * (1-est) / n) ))
```
jgabry, line 176:
```suggestion
  list(estimate = est, se = sqrt(est * (1-est) / n) )
```

R/loo_predictive_error.R
```diff
@@ -0,0 +1,200 @@
...
+  yhat <- as.integer(yhat > 0.5)
+  mask <- y == 0
+
+  tn <- mean(yhat[mask] == y[mask]) # True negatives
+  tp <- mean(yhat[!mask] == y[!mask]) # True positives
+
+  bls_acc <- (tp + tn) / 2
+  # This approximation has quite large bias for small samples
+  bls_acc_var <- (tp * (1 - tp) + tn * (1 - tn)) / 4
+  return(list(estimate = bls_acc, se = sqrt(bls_acc_var / n)))
```
jgabry, line 198:
```suggestion
  list(estimate = bls_acc, se = sqrt(bls_acc_var / n))
```

R/loo_predictive_error.R
```diff
@@ -0,0 +1,200 @@
...
+loo_predictive_error <- function(x, ...) {
+  UseMethod("loo_predictive_error")
+}
+
+#' @rdname loo_predictive_error
+#' @export
+loo_predictive_error.default <-
+  function(x,
+           y,
+           ll,
```
jgabry, line 87: I wonder if we should call the argument `log_lik` instead of `ll` since for `psis()` we have an argument `log_ratios` and not `lr`. Not a big deal either way. @avehtari What do you think?

R/loo_predictive_error.R
```diff
@@ -0,0 +1,200 @@
...
+           pred_error = c("mae", "rmse", "mse", "acc", "balanced_acc"),
```
jgabry, line 88: What do you (and @avehtari) think about the name `metric` instead of `pred_error`?

jgabry (commented):
This looks great, thanks! I made a bunch of comments/suggestions but they're all really minor.

LeeviLindgren:
> @LeeviLindgren Thanks for the PR! Sorry for the delay in reviewing it. I'll make a few review comments now, but can you also edits from maintainers? You can do that by following the instructions here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

Thanks for the comments, I will take a look! Edits from maintainers should be allowed.

LeeviLindgren:
@avehtari also suggested that it would be nice to support model comparison using these additional metrics. I suggest we try to merge this PR which implements the metrics, and I can start working on the functionality to support comparison. The initial idea I had was to write a method for `loo_compare` for `loo_predictive_error`-class (we can also come up with another name). What do you think @jgabry?

jgabry:
Thanks for the updates. Waiting to hear from @avehtari on the couple of questions in my review comments (about the argument names) but after that we can merge this (assuming the GitHub Actions checks pass). 

> @avehtari also suggested that it would be nice to support model comparison using these additional metrics. I suggest we try to merge this PR which implements the metrics, and I can start working on the functionality to support comparison. The initial idea I had was to write a method for `loo_compare` for `loo_predictive_error`-class (we can also come up with another name). Users could then pass output from the calls to `loo_predictive_error` to `loo_compare`. What do you think @jgabry?

Yeah I agree it would be nice to be able to do model comparison with these. My initial thought is that your suggestion of using a class and a `loo_compare` method seems like a good approach.

R/loo_predictive_error.R
avehtari: longer name is more descriptive, and thus better

R/loo_predictive_error.R
avehtari: Good idea, I like `metric`, especially as "acc" is not an error

avehtari:
I agree with @jgabry's suggestions

R/loo_predictive_error.R
LeeviLindgren: I think it would also then make sense to call the function `loo_predictive_metric`, any opinions on this?

R/loo_predictive_error.R
jgabry: Ok let's go with `log_lik`. @LeeviLindgren Can you update the argument name?

R/loo_predictive_error.R
jgabry: Yeah I suppose if we're including metrics that aren't errors then `loo_predictive_error` isn't the best name.

R/loo_predictive_error.R
```diff
@@ -0,0 +1,202 @@
...
+#' # data from help("lm")
+#' ctl <- c(4.17,5.58,5.18,6.11,4.50,4.61,5.17,4.53,5.33,5.14)
+#' trt <- c(4.81,4.17,4.41,3.59,5.87,3.83,6.03,4.89,4.32,4.69)
+#' d <- data.frame(
+#'   weight = c(ctl, trt),
+#'   group = gl(2, 10, 20, labels = c("Ctl","Trt"))
+#' )
+#' fit <- stan_glm(weight ~ group, data = d, refresh = 0)
+#' ll <- log_lik(fit)
+#' r_eff <- relative_eff(exp(-ll), chain_id = rep(1:4, each = 1000))
```
jgabry, line 68: I think the doc needs to be regenerated one more time so that the .Rd file includes this change to `-ll` instead of `log_ratios`. That's why R cmd check is failing but the unit tests are passing.

LeeviLindgren:
- I changed the function to `loo_predictive_metric`
- Updated the wording in docs to match the updates
- `ll` argument is now called `log_lik`
- `pred_error` argument is now called `metric`

jgabry:
Looks good, thanks! Merging now.

Issue #201, opened 2022-05-10
Allow use of other criteria


sims1253:
It would be nice if the loo objects weren't hardcoded to use elpd. For a study I wanted to include rmse and R2 and had to build fake loo objects in order to use the loo_compare functionality. Code is [here](https://github.com/sims1253/bayesim/blob/master/R/loo_handler.R).

avehtari:
@sims1253 do you need something else than what is in the PRs mentioned above, or can we close this issue?

avehtari:
We do have rmse, but we're still missing R2 (which is available in brms and rstanarm, but not as plain loo package version)

sims1253:
> ...which is available in brms

At least the brms version does not use psis, which is why I wrote the version from my pr.

avehtari:
brms `loo_R2` does use psis based on the code https://github.com/paul-buerkner/brms/blob/d1a050803670f4f4647324ce9a84780591a40692/R/loo_predict.R#L202

sims1253:
True. Sorry, I'm currently on vacation and only have my phone and only found the Bayes r2 function. 
Maybe it was about the pointwise estimate for the SE then that we built the new version. I don't quite remember to be honest.

PR #203, opened 2022-06-15
CRPS and SCRPS

LeeviLindgren:
Addresses #136. Adds functions to compute `crps` and `scrps` and their `loo` versions. Happy to hear feedback!

PR #205, opened 2022-10-28
Add R2

sims1253:
This is based on the code I wrote for bayesim. Related issue #201


jgabry:
Thanks for the PR, will try to review soon! 

@avehtari what do you think about adding this?

jgabry:
@sims1253 Can you pull in the latest changes from the master branch into your PR (I updated the GitHub actions workflow files on the master branch) so that it runs the tests again? Thanks

sims1253:
Forgot to push the reference values for the tests >.> Lost them in all the styler changes I didn't want to add to this pr.
I am not too familiar with the whole R generics thing and what the plan is for loo so I guess that might need some touching.

avehtari:
> @avehtari what do you think about adding this?

I'm in favor of adding R2. We need to be careful that all these additional utility/loss functions (see #202 and #203) have the same behavior for the user.

Based on a quick look, the SE is not computed correctly, as it's ignoring the uncertainty in ss_y. If I remember correctly, we did go through the correct derivation with @LeeviLindgren in summer?

LeeviLindgren:
@avehtari We derived the first-order Taylor approximation of the variance. What we came up with is 
image, where $MSE_y$ and $MSE_{\hat{e}}$ are the mean squared errors of predicting data with its mean and  LOO-predictions, respectively (note that the equality sign in the image should be $\approx$).

sims1253:
Thanks for the feedback. If my memory serves me right, we defined R2 in a slightly different way than what you used in your [case study](https://avehtari.github.io/bayes_R2/bayes_R2.html) to get the pointwise entries for psis-loo and the SE. But this was a year ago so I need to go check my notes to see what I thought back then and won't have time this week.

avehtari:
@sims1253, I'm just checking whether you have had time to check your notes?

avehtari:
@sims1253 any update on this?

sims1253:
I assumed that I could use the same formula as in `table_of_estimates` to get the SE from the pointwise R2. Iirc this was one of the reasons we used the rmse inspired R2 = 1 - SS_e/SS_y approach. 

This is from my notes:
Starting with the RMSE as 
$$RMSE = \sqrt{\frac{1}{N}\sum_{n=1}^{N}(y_n - \hat{y}_n)^2}$$

We integrate/mean over the posterior samples (and ignore the sum and 1/N until later)
$$\int...\int\sqrt{\frac{1}{N}\sum_{n=1}^N(y_n - \hat{y}_n)^2} p(\hat{y}_1) d\hat{y}_1...p(\hat{y}_S)d\hat{y}_S$$

Which we can factorize into
$$\int...\int (y_1 - \hat{y}_1)^2 + ... + (y_N - \hat{y}_N)^2 p(\hat{y}_1) d\hat{y}_1...p(\hat{y}_S)d\hat{y}_S$$

and reform into
$$\int...\int (y_1 - \hat{y}_1)^2 p(\hat{y}_1) d\hat{y}_1...p(\hat{y}_S)d\hat{y}_S + ... + \int\int (y_N - \hat{y}_N)^2 p(\hat{y}_1) d\hat{y}_1...p(\hat{y}_S)d\hat{y}_S$$

If the s's in the $(y_s - \hat{y}_s)^2$ and $p(\hat{y}_s)d\hat{y}_s$ are not the same, they're constant so we can pull them out.
Pulling the  $(y_s - \hat{y}_s)^2$ terms out of the integral leaves $S$ $\int p(\hat{y}_s)d\hat{y}_s$ integrals that all equal 1as they are integrals over a density.

I am missing a step here in my notes but hopefully you are a lot better with integrals than I am. 

$$\sum_{n=1}^N \int (y_n - \hat{y}_n)^2p(\hat{y}_n)d\hat{y}_n$$

Adding back the sum and 1/N gives us
 $$\frac{1}{N}\sum_{n=1}^N \sqrt{\int (y_n - \hat{y}_n)^2p(\hat{y}_n)d\hat{y}_n}$$

 which is defined pointwise and integrates over posterior samples. So we can apply psis weights and have the benefit of having a pointwise result for SE estimation.


(I am sorry but I can't find the reason for these not to render properly)

Moving on to R2, we use the following definition of R2
$$R^2 = 1 - \frac{SS_e}{SS_y} = 1 - \frac{\sum_{n = 1}^N (y_n - \hat{y}_n)^2}{\sum_{n = 1}^N (y_n - \bar{y})^2}$$

As there are not posterior samples in the denominator the integral form looks like this:

$$\int R^2 dp(\hat{y}) = 1 - \frac{\sum_{n=1}^N \int (y_n - \hat{y}_n)^2 dp(\hat{y}_n)}{\sum_{n=1}^N (y_n - \bar{y})^2}$$

$$= \sum_{n = 1}^N ( \frac{1}{N} - \frac{\int (y_n - \hat{y}_n)^2}{\sum_{k = 1}^N (y_k - \bar{y})^2})$$

For psis we only have to weight the nominator and again we have a pointwise result.

sims1253:
Ah I think I understand the problem now, thanks for sticking with me. Will try to update this PR over the holidays.

sims1253:
Closing this for now so it doesn't clutter your PR list.

fweber144:
@LeeviLindgren sorry for digging this out; it's because of https://github.com/stan-dev/projpred/pull/496. Do I understand

> We derived the first-order Taylor approximation of the variance.

in https://github.com/stan-dev/loo/pull/205#issuecomment-1316683962 correctly that you applied a [bivariate delta method](https://en.wikipedia.org/wiki/Delta_method#Multivariate_delta_method) to the function $R^2(MSE_{\hat{e}}, MSE_{y}) = 1 - \frac{MSE_{\hat{e}}}{MSE_{y}}$ where both $MSE_{\hat{e}}$ and $MSE_{y}$ are random variables?

fweber144:
> @LeeviLindgren sorry for digging this out; it's because of [stan-dev/projpred#496](https://github.com/stan-dev/projpred/pull/496). Do I understand
> 
> > We derived the first-order Taylor approximation of the variance.
> 
> in [#205 (comment)](https://github.com/stan-dev/loo/pull/205#issuecomment-1316683962) correctly that you applied a [bivariate delta method](https://en.wikipedia.org/wiki/Delta_method#Multivariate_delta_method) to the function R 2 ( M S E e ^ , M S E y ) = 1 − M S E e ^ M S E y where both M S E e ^ and M S E y are random variables?

I now did the math myself and I obtain the same result when proceeding as I described. So this result comes indeed from a bivariate delta method applied to that $R^2(\cdot, \cdot)$ function.

PR #210, opened 2022-12-27
MixIS vignette by Luca Silva and Giacomo Zanella

avehtari:
Luca Silva and Giacomo Zanella created a vignette, which I slightly edited and added data to the package. The vignette demonstrates the mixture IS approach by [Silva and Zanella (2022)](https://arxiv.org/abs/2209.09190). 

I hope the vignette is not too slow and with too big data?


vignettes/loo2-mixis.Rmd
```diff
@@ -0,0 +1,198 @@
...
+```{r}
+data(voice_loo)
+elpd_loo <- voice_loo$elpd_loo
+```
+
+We can then compute the root mean squared error (RMSE) of the PSIS and mixture estimators relative to such benchmark values.
+```{r}
+elpd_psis <- loo_post$pointwise[,1]
+print(paste("RMSE(PSIS) =",round( sqrt(mean((elpd_loo-elpd_psis)^2)) ,2)))
+print(paste("RMSE(MixIS) =",round( sqrt(mean((elpd_loo-elpd_mixis)^2)) ,2)))
```
jgabry, line 173: When I run this these come out as NaN. It looks like `voice_loo` has columns "X" and "bk_loo" but no column `elpd_loo`. So `voice_loo$elpd_loo` comes out to be NULL.

jgabry (changes requested):
This looks good, except for the one NaN issue I mentioned in my other comment

jgabry:
> I hope the vignette is not too slow and with too big data?

I think the speed and data size are ok, so just the one problem I pointed out above

vignettes/loo2-mixis.Rmd
avehtari: should be fixed now

jgabry: Thanks! I'll take a look soon

jgabry:
@avehtari This is now online. Here's the URL for the vignette: https://mc-stan.org/loo/articles/loo2-mixis.html

PR #217, opened 2023-03-25
fix loo_model_weights naming for list of loo objects

### LINKED ISSUE #165: loo_compare and loo_model_weights have different behavior for list argument
If the list of loo objects has some named elements the output column names behave differently. For example,
```
> loo_compare(list(loo1,`Model f1+f2`=loo2))
            elpd_diff se_diff
Model f1+f2    0.0       0.0 
model1      -530.0      31.1 

> loo_model_weights(list(loo1,`Model f1+f2`=loo2))
Method: stacking
------
       weight
model1 0.000 
model2 1.000 
```
If all list elements are named, then both behave nicely
```
> loo_compare(list(`Model f1`=loo1,`Model f2`=loo2))
            elpd_diff se_diff
Model f1+f2    0.0       0.0 
Model f1    -530.0      31.1 
> loo_model_weights(list(`Model f1`=loo1,`Model f1+f2`=loo2))
Method: stacking
------
            weight
Model f1    0.000 
Model f1+f2 1.000 
```

It would be good to fix this, and then demonstrate the use of naming models in vignettes and help text, as people sometimes get confused what is the order as the order is not the same as the order of loo objects in the argument and the order can be different in `loo_compare` and in `loo_model_weights`.

jgabry:
fixes #165 by getting model names for `loo_model_weights` even if list of loo objects is only partially named 

@avehtari Does this behave as expected now?

### DIFF
--- a/R/loo_compare.R
+++ b/R/loo_compare.R
@@ -231,8 +231,7 @@ find_model_names <- function(x) {
       out_names[j] <- names4[j]
     }
   }
-
-  return(out_names)
+  out_names
 }
 
 
--- a/R/loo_model_weights.R
+++ b/R/loo_model_weights.R
@@ -6,16 +6,16 @@
 #' Gelman, Yao, and Gabry (2019) for background.
 #'
 #' @export
-#' @param x A list of pointwise log-likelihood matrices or `"psis_loo"` objects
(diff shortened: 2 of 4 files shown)


avehtari (approved):
Tested. Works now as expected.

Issue #222, opened 2023-04-11
Adding "loo" criterion with add_criterion and moment_match=TRUE is failing even when save_pars(all=TRUE) was set during model fit


eipi10:
I'm fitting ordinal regression models using brms, but I'm getting an error when I try to add a loo criterion with `moment_match=TRUE`, even though I fit the model with `save_pars(all=TRUE)`. The error message is `Error in if (quantities_i$ki < ki) { : missing value where TRUE/FALSE needed.` 

I've created some fake data with a similar structure to my real data (but with far fewer observations and with only a few of the columns that are in my real data) in order to reproduce the error. Here is a link to the sample data:
 
[sample-data.csv](https://github.com/paul-buerkner/brms/files/11181049/sample-data.csv)

Even this sample has a relatively large number of rows (20,591) as I wasn't able to reproduce the error with smaller data frames. 

Reproducible code is below, along with the warnings and error messages. 

(FYI, I [posted a question](https://discourse.mc-stan.org/t/add-criterion-with-moment-match-true-failing-even-when-save-pars-all-true-was-set-during-model-fit/30956) about this on the Stan Forum several days ago but didn't get an answer. I also originally posted this issue on the [`brms` repo](https://github.com/paul-buerkner/brms/issues/1479), but Paul pointed out that the error is coming from `loo` and asked me to post the issue here.)

```
# Session Info
# Macbook Pro M1 Max, Ventura 13.2.1
# R 4.2.3
# brms 2.19.0
# rstan 2.26.13
# stanheaders 2.26.13
# cmdstanr 0.5.2
# cmdstan 2.31.0
# loo 2.6.0

library(tidyverse)
library(brms)

dd = read.csv("sample-data.csv") %>%
  mutate(student.id = factor(student.id),
         instructor.id = factor(instructor.id),
         grade = ordered(grade, levels=c("F","D-","D","D+","C-","C","C+","B-","B","B+","A-","A")))

form = grade ~ incoming.gpa + learning.mode + instructor.rank +
  (1|instructor.id) + (1|student.id)

priors = c(prior(normal(0,0.5), class="b"),
           prior(normal(0,1), class="b", coef="incoming.gpa"))

mod1 = brm(data=dd, 
           file="models/mod1",
           formula=form, 
           family=cumulative(link="logit", threshold="flexible"),
           prior=priors,
           chains=3, cores=3, threads=threading(3),
           backend="cmdstanr",
           sample_prior=TRUE,
[...]

PR #224, opened 2023-07-01
make Pareto k Inf if it is NA

topipa:
This PR changes the handling of Pareto k values in edge cases. If Pareto k is NaN or NA, it will be changed to Inf, whereas previously it was done only for NaN. Based on the discussion in https://github.com/stan-dev/loo/issues/222.


jgabry:
Thanks @topipa!

PR #226, opened 2023-07-27
Fix wvar

### LINKED ISSUE #225: `.wvar` internal function isn't calculating the right thing
See @charlesm93's post and @avehtari's reply on the forum: https://discourse.mc-stan.org/t/e-loo-function-in-loo-package-type-var/32151/6

jgabry:
Fixes #225 

Fixes internal .wvar computation and adds std deviation option


jgabry:
The checks are failing because I need to regenerate the documentation (will do that a little later today), but all tests are passing.

jgabry:
Ok everything is passing now. @avehtari this is ready for a quick review whenever you have time.

R/E_loo.R
```diff
@@ -199,12 +193,11 @@ E_loo.matrix <-
...
   sum(w * x)
 }
-.wvar <- function(x, w, r_eff = NULL, ...) {
-  if (is.null(r_eff)) {
-    r_eff <- 1
-  }
-  r <- (x - .wmean(x, w))^2
-  sum(w^2 * r) / r_eff
+.wvar <- function(x, w, ...) {
+  (sum(.wmean(x^2, w)) - sum(.wmean(x, w)^2)) / (1 - 1/length(w))
```
avehtari, line 197: While thinking a test with non-equal w's, I realized that here `length(w)` should be replaced with the effective sample size `1/sum(w^2)`. To make it clear, maybe write
```
ess <- 1/sum(w.^2);
(sum(.wmean(x^2, w)) - sum(.wmean(x, w)^2)) / (1 - 1/ess)
```

To see why we need to use ess, consider `w <- c(rep(0.1,10), rep(0, 90))`, which has 10 non-zero but equal weights, and the correct variance would be `var(x[w>0])`. Using the ESS estimate `1/sum(w^2)` works correctly then also when all weights are non-zero bu not equal.

avehtari: I also first thought proposing `(1- sum(w^2))` but maybe i can be used with a comment like
```
# (1- sum(w^2)) is equal to (ESS-1)/ESS, where effective sample size ESS is estimated with the generic target quantity invariant estimate 1/sum(w^2), see e.g. "Monte Carlo theory, methods and examples" by Owen (2013).

jgabry: Ok the `.wvar` function now has denominator `(1 - sum(w^2))` and the comment you suggested adding. Can you take one more quick look and then if everything looks good we can merge this. 

I will also run reverse dependency checks to make sure that none of the packages that depend on loo have errors resulting from this change (for example if they have a test in their package using `E_loo(type="var")` then the result will change). If there are errors I will notify maintainers of the upcoming change.

avehtari: Looks good now, and I also tested with a slightly different example that the computation matches var

avehtari (approved):
ok

R/E_loo.R
jgabry: > I will also run reverse dependency checks to make sure that none of the packages that depend on loo have errors resulting from this change (for example if they have a test in their package using E_loo(type="var") then the result will change).

No issues with reverse dependencies

PR #230, opened 2023-09-29
Add order statistic warning

yannmclatchie:
This PR adds a warning when we estimate that the elpd difference of a collection of models is purely due to chance, developed with @avehtari.

If more than 11 models are compared, then the median model by elpd is taken as the baseline model, and the risk of the difference in predictive performance being due to random noise is estimated as described by McLatchie and Vehtari (Section 3.2, [2023](https://arxiv.org/abs/2309.03742)). This will flag a warning if it is deemed that there is a risk of over-fitting due to the selection process, and users are recommended to avoid model selection based on LOO-CV, and instead to favour of model averaging/stacking or projection predictive inference.

### DIFF
--- a/R/loo_compare.R
+++ b/R/loo_compare.R
@@ -44,7 +44,7 @@
 #'
 #'   If more than \eqn{11} models are compared, then the worst model by elpd is
 #'   taken as the baseline model, and the risk of the difference in predictive
-#'   performance being due to random noise is estimated as described by
+#'   performance being due to chance is estimated as described by
 #'   McLatchie and Vehtari (2023). This will flag a warning if it is deemed that
 #'   there is a risk of over-fitting due to the selection process, and users
 #'   are recommended to avoid model selection based on LOO-CV, and
@@ -319,7 +319,7 @@ loo_order_stat_check <- function(loos, ord) {
 
   if (max(elpd_diff) <= order_stat) {
     # flag warning if we suspect no model is theoretically better than the baseline
-    warning("Difference in performance potentially due to random noise.",
+    warning("Difference in performance potentially due to chance.",
             call. = FALSE)
   }
 }
--- a/R/loo_compare.R
+++ b/R/loo_compare.R
@@ -42,16 +42,14 @@
 #'   distribution, a practice derived for Gaussian linear models or
 #'   asymptotically, and which only applies to nested models in any case.
 #'
-#'   If more than \eqn{11} models are compared, then the median model by elpd is
-#'   taken as the baseline model, and we recompute (internally) the model
-#'   differences to this baseline. We then estimate whether the difference in
-#'   predictive performances is potentially due to chance as described by
-#'   McLatchie and Vehtari (2023). This will flag a warning if it is deemed that
-#'   there is a risk of over-fitting due to the selection process, and users
-#'   are recommended to avoid model selection based on LOO-CV, and
-#'   instead to favour of model averaging/stacking or projection predictive
-#'   inference.
-#'
+#'   If more than \eqn{11} models are compared, we internally recompute the model
+#'   differences using the median model by ELPD as the baseline model. We then
+#'   estimate whether the differences in predictive performance are potentially
+#'   due to chance as described by McLatchie and Vehtari (2023). This will flag
+#'   a warning if it is deemed that there is a risk of over-fitting due to the
+#'   selection process. In that case users are recommended to avoid model
+#'   selection based on LOO-CV, and instead to favor model averaging/stacking or
+#'   projection predictive inference.
 #' @seealso
 #' * The [FAQ page](https://mc-stan.org/loo/articles/online-only/faq.html) on
 #'   the __loo__ website for answers to frequently asked questions.
--- a/R/loo_compare.R
+++ b/R/loo_compare.R
@@ -43,11 +43,11 @@
 #'   asymptotically, and which only applies to nested models in any case.
 #'
 #'   If more than \eqn{11} models are compared, then the worst model by elpd is
-#'   taken as the baseline model, and the risk of difference in predictive
+#'   taken as the baseline model, and the risk of the difference in predictive
 #'   performance being due to random noise is estimated as described by
 #'   McLatchie and Vehtari (2023). This will flag a warning if it is deemed that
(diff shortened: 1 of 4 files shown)


R/loo_compare.R
```diff
@@ -270,3 +282,55 @@ loo_compare_order <- function(loos){
...
+    # procedure cannot be diagnosed for fewer than ten candidate models
+    # (total models = worst model + ten candidates)
+    # break from function
+    return(NULL)
+  }
+
+  ## warnings
+
+  # compute the elpd differences
+  worst_idx <- tail(ord, n=1)
```
jgabry, line 306:
```suggestion
  worst_idx <- utils::tail(ord, n=1)
```

R/loo_compare.R
```diff
@@ -270,3 +282,55 @@ loo_compare_order <- function(loos){
...
+  diffs <- mapply(FUN = elpd_diffs, loos[ord[worst_idx]], loos[ord])
+  elpd_diff <- apply(diffs, 2, sum)
+
+  # estimate the standard deviation of the upper-half-normal
+  diff_median <- median(elpd_diff)
```
jgabry, line 311:
```suggestion
  diff_median <- stats::median(elpd_diff)
```

R/loo_compare.R
jgabry: This will fix the R cmd check NOTE in https://github.com/stan-dev/loo/actions/runs/6350707007/job/17258966102?pr=230#step:6:96 about no visible global function definition

R/loo_compare.R
jgabry: This will fix the R cmd check NOTE in https://github.com/stan-dev/loo/actions/runs/6350707007/job/17258966102?pr=230#step:6:96 about no visible global function definition

R/loo_compare.R
```diff
@@ -270,3 +282,55 @@ loo_compare_order <- function(loos){
...
+  n_models <- sum(!is.na(elpd_diff_trunc))
+  candidate_sd <- sqrt(1 / n_models * sum(elpd_diff_trunc^2, na.rm = TRUE))
+
+  # estimate expected best diff under null hypothesis
+  K <- length(loos) - 1
+  order_stat <- order_stat_heuristic(K, candidate_sd)
+
+  if (max(elpd_diff) <= order_stat) {
+    # flag warning if we suspect no model is theoretically better than the baseline
+    warning("Difference in performance potentially due to chance.",
```
jgabry, line 322: Should we point users to the paper in the warning message (e.g., `"See McLatchie and Vehtari (2023) for details".`)? @avehtari what do you think?

R/loo_compare.R
```diff
@@ -42,10 +42,19 @@
 #'   distribution, a practice derived for Gaussian linear models or
 #'   asymptotically, and which only applies to nested models in any case.
 #'
+#'   If more than \eqn{11} models are compared, then the worst model by elpd is
+#'   taken as the baseline model, and the risk of the difference in predictive
```
jgabry, line 46: I'm slightly confused by this because the printed `loo_compare` results will still show the model with the _best_ ELPD as the baseline (i.e. all `elpd_diff` are relative to that one), right? 

If that's still true, then I think we should rephrase this to indicate that this is just happening in an internal check and doesn't affect which model is considered as the baseline for computing `elpd_diff` in the output. Does that make sense? Or am I just confused (totally possible)?

jgabry (commented):
@yannmclatchie Great, thank you for the PR! I put a few suggestions and questions in the review comments.

R/loo_compare.R
yannmclatchie: I have now updated the docstring, let me know if it is now clearer 🙏

yannmclatchie:
Thanks for the feedback @jgabry ! I've gone through your suggestions and made the changes with also a small change in the logic: we use the median model rather than the worst model to take the _internal_ differences and compute the order statistic.

I have now also updated the documentation to make it clear that this second layer of elpd differences is different to what the user will eventually see and is strictly internal, as you noted. Let me know if there are any further changes that could be made to clarify

R/loo_compare.R
avehtari: yes

yannmclatchie: I have now added this

R/loo_compare.R
```diff
@@ -42,10 +42,20 @@
...
+#'   If more than \eqn{11} models are compared, then the median model by elpd is
+#'   taken as the baseline model, and we recompute (internally) the model
+#'   differences to this baseline. We then estimate whether the difference in
+#'   predictive performances is potentially due to chance as described by
+#'   McLatchie and Vehtari (2023). This will flag a warning if it is deemed that
+#'   there is a risk of over-fitting due to the selection process, and users
+#'   are recommended to avoid model selection based on LOO-CV, and
+#'   instead to favour of model averaging/stacking or projection predictive
+#'   inference.
+#'
```
jgabry, line 54:
```suggestion
#'   If more than \eqn{11} models are compared, we internally recompute the model
#'   differences using the median model by ELPD as the baseline model. We then
#'   estimate whether the differences in predictive performance are potentially
#'   due to chance as described by McLatchie and Vehtari (2023). This will flag
#'   a warning if it is deemed that there is a risk of over-fitting due to the
#'   selection process. In that case users are recommended to avoid model
#'   selection based on LOO-CV, and instead to favor model averaging/stacking or
#'   projection predictive inference.
```

jgabry (changes requested):
@yannmclatchie Sorry for the delay on this! I've been super busy recently. I added one more comment suggesting minor edits to the documentation, but otherwise this looks good. After you take a look at my suggestion and approve it I can merge this PR!

@avehtari Does this look good to you too?

jgabry:
Thanks for updating the doc @yannmclatchie (and thanks for adding this to the package!). @avehtari if you're happy with this I think we can go ahead and merge it once all the automated tests are run again.

avehtari:
I can check this at latest on Tuesday (too late today)

jgabry:
Sounds good, thanks

avehtari:
OK for me

jgabry (approved):
Merging now. Thanks again!

PR #241, opened 2024-02-02
pointwise() convenience function for extracting pointwise estimates

### LINKED ISSUE #126: Convenience functions for accessing pointwise estimates
Right now to get the pointwise values for elpd_loo, p_loo, etc., you have to do this: 

```r
object$pointwise[, "elpd_loo"]
```

but it would be nice if we offered a convenience function for this, e.g., 

```r
pointwise(object, "elpd")
```
which would return a vector of pointwise elpd values.

jgabry:
Closes #126

I ended up wanting this to make some code look nicer and it was very easy to implement, so I made a PR. 

```r
pointwise(loo_object, "elpd_loo") # returns vector of pointwise elpd_loo estimates
pointwise(loo_object, "p_loo") # returns vector of pointwise p_loo estimates
# etc. 
```

This works not only for `loo()` but also for `loo_subsample()`, `loo_moment_match()`, etc.


jgabry:
@avehtari This is super simple, so you don't need to spend much (or any) time looking at this, but just wanted to check if you're ok with adding this little convenience function. I think it's ready to merge if it's ok with you. I'm starting to prepare the release.

avehtari (approved):
This makes sense. Looks good

PR #246, opened 2024-02-15
simplify mcse_elpd using log-normal approximation

avehtari:
Years ago I made mcse_elpd using a simple quadrature, but now switched to use a more efficient log-normal approximation. Given the mean and variance in linear scale, and knowing that the logarithm is closer to normal, we approximate the uncertainty in linear scale with log-normal and compute the corresponding normal distribution parameters in log scale using the equations shown, e.g., at https://en.wikipedia.org/wiki/Log-normal_distribution#Arithmetic_moments

The change makes relative change of 0.01%-0.1% compared to stored reference values and thus some reference rds files were updated.


jgabry (approved):
Looks good. I'll include this in the release.

PR #251, opened 2024-02-29
make E_loo Pareto-k diagnostic more robust

### LINKED ISSUE #250: E_loo fails with github version of posterior
The latest `loo` release uses `posterior::pareto_khat()` as in the latest `posterior` CRAN release. The github version of `posterior` has breaking changes. I can fix this tomorrow, but creating the issue now. Ping @n-kall, too

avehtari:
Fixes #250 (although the reason was not posterior version)

E_loo was giving warnings and errors in certain cases with binary x. If h(theta) is a step function and h is 0/1 binary, then when computing Pareto-k the one of the tails is likely to be just 0's. In such cases, using the same argument as for type="quantile", we can return the Pareto-k just for r. To make the function to complain less, the same approach is now used for constant, NA, NaN, and infinite x. Added also tests that these don't cause warnings.

### DIFF
--- a/R/E_loo.R
+++ b/R/E_loo.R
@@ -48,11 +48,11 @@
 #'   Pareto-k's, which may produce optimistic estimates.
 #'
 #'   For `type="mean"`, `type="var"`, and `type="sd"`, the returned Pareto-k is
-#'   the maximum of the Pareto-k's for the left and right tail of \eqn{hr} and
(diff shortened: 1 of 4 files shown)


jgabry (approved):
Looks good. I fixed a typo in a comment so the unit tests are running again but this is ready to merge when they finish running.

PR #259, opened 2024-03-04
fix loo_moment_matching NaN issue

### LINKED ISSUE #258: loo_moment_match fails if new log likelihood is -Inf
I had seen the same issue in some of my models, but started to investigate after JackL posted a simple example in Stan discourse
https://discourse.mc-stan.org/t/error-when-adding-looic-to-a-zero-one-inflated-beta-regression-model-in-brms/34337

The problem arises when `log_liki_new` and `log_prob_new` both evaluate as -Inf, and the log ratio contains `-log_liki_new + log_prob_new` and -Inf + Inf returns NaN. The similar computation can happen in split moment matching.

In these cases, it would be fine that we reject the problematic evaluation in the same way as Stan would reject it during sampling. In importance sampling we can do rejection by making the weight to be 0, which me can make by making the log ratio -Inf (thus we now compute -Inf + Inf is -Inf). We can safely allow -Infs in our importance sampling algorithms.

avehtari:
Fixes #258 

- In `loo_moment_matching` and `split_moment_matching` replace NaN's in log ratios resulting from -Inf + Inf with -Inf before calling `importance_sampling.default()`
- in `validate_ll()` allow -Inf
- update tests

### DIFF
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -57,7 +57,7 @@ validate_ll <- function(x) {
     stop("List not allowed as input.")
   } else if (anyNA(x)) {
     stop("NAs not allowed in input.")
-  } else if (any(x==Inf)) {
+  } else if (any(x == Inf)) {
     stop("All input values must be finite or -Inf.")
   }
   invisible(x)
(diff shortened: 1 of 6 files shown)


jgabry (approved):
Looks good

PR #262, opened 2024-03-19
catch Stan log_prob exceptions inside moment matching

### LINKED ISSUE #261: Stan log_prob exceptions cause moment_matching to fail
Reported in Stan discourse with reproducible example
https://discourse.mc-stan.org/t/add-criterion-with-moment-match-true-failing-even-when-save-pars-all-true-was-set-during-model-fit/30956/13

The moment matching transformed unconstrained parameter values can under- or overflow when constrained or some further quantity in the Stan program can under- or overflow. In this example one of the variables inside the model is 0, while should be positive. This causes an exception, which causes an error in R and moment_matching quits:
```
Error : Exception: beta_lpdf: Second shape parameter is 0, but must be positive finite! (in 'string', line 226, column 6 to column 111)
```

`brms` catches this error, but makes only one guess for the reason
```
Error: Moment matching failed. Perhaps you did not set 'save_pars = save_pars(all = TRUE)' when fitting your model? If you are running moment matching on another machine than the one used to fit the model, you may need to set recompile = TRUE.
```

avehtari:
Fixes #261

When doing moment matching steps, catch the errors when evaluating log_prob, and break out without an error. This will allow getting moment matching results for all the non-failing observations and may get some moment matching also for the failing ones.

### DIFF
--- a/R/loo_moment_matching.R
+++ b/R/loo_moment_matching.R
@@ -286,7 +286,7 @@ loo_moment_match_i <- function(i,
     # 1. match means
     trans <- shift(x, uparsi, lwi)
     # gather updated quantities
-    quantities_i <- tryCatch(update_quantities_i(x, trans$upars,  i = i,
+    quantities_i <- try(update_quantities_i(x, trans$upars,  i = i,
                                         orig_log_prob = orig_log_prob,
                                         log_prob_upars = log_prob_upars,
                                         log_lik_i_upars = log_lik_i_upars,
@@ -295,7 +295,7 @@ loo_moment_match_i <- function(i,
                                         is_method = is_method,
                                         ...)
                         )
-    if (is_try_error(quantities_i)) {
+    if (inherits(quantities_i, "try-error")) {
       # Stan log prob caused an exception probably due to under- or
       # overflow of parameters to invalid values
       break
@@ -325,7 +325,7 @@ loo_moment_match_i <- function(i,
                                         is_method = is_method,
                                         ...)
                         )
-    if (is_try_error(quantities_i)) {
+    if (inherits(quantities_i, "try-error")) {
       # Stan log prob caused an exception probably due to under- or
       # overflow of parameters to invalid values
       break
@@ -358,7 +358,7 @@ loo_moment_match_i <- function(i,
                                           ...)
                           )
 
-      if (is_try_error(quantities_i)) {
+      if (inherits(quantities_i, "try-error")) {
         # Stan log prob caused an exception probably due to under- or
         # overflow of parameters to invalid values
         break
--- a/R/loo_moment_matching.R
+++ b/R/loo_moment_matching.R
@@ -286,7 +286,7 @@ loo_moment_match_i <- function(i,
     # 1. match means
     trans <- shift(x, uparsi, lwi)
     # gather updated quantities
-    quantities_i <- update_quantities_i(x, trans$upars,  i = i,
+    quantities_i <- tryCatch(update_quantities_i(x, trans$upars,  i = i,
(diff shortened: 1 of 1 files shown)


R/loo_moment_matching.R
```diff
@@ -336,15 +348,21 @@ loo_moment_match_i <- function(i,
...
                                           orig_log_prob = orig_log_prob,
                                           log_prob_upars = log_prob_upars,
                                           log_lik_i_upars = log_lik_i_upars,
                                           r_eff_i = r_eff_i,
                                           cores = 1,
                                           is_method = is_method,
                                           ...)
+                          )
 
+      if (is_try_error(quantities_i)) {
```
jgabry, line 361:
```suggestion
      if (inherits(quantities_i, "try-error")) {
```

R/loo_moment_matching.R
```diff
@@ -310,14 +316,20 @@ loo_moment_match_i <- function(i,
...
+    quantities_i <- try(update_quantities_i(x, trans$upars,  i = i,
                                         orig_log_prob = orig_log_prob,
                                         log_prob_upars = log_prob_upars,
                                         log_lik_i_upars = log_lik_i_upars,
                                         r_eff_i = r_eff_i,
                                         cores = 1,
                                         is_method = is_method,
                                         ...)
+                        )
+    if (is_try_error(quantities_i)) {
```
jgabry, line 328:
```suggestion
    if (inherits(quantities_i, "try-error")) {
```

R/loo_moment_matching.R
```diff
@@ -286,14 +286,20 @@ loo_moment_match_i <- function(i,
...
+    quantities_i <- tryCatch(update_quantities_i(x, trans$upars,  i = i,
                                         orig_log_prob = orig_log_prob,
                                         log_prob_upars = log_prob_upars,
                                         log_lik_i_upars = log_lik_i_upars,
                                         r_eff_i = r_eff_i,
                                         cores = 1,
                                         is_method = is_method,
                                         ...)
+                        )
+    if (is_try_error(quantities_i)) {
```
jgabry, line 298: I don't think `is_try_error` is a base R function. Is this from a package? I think not finding this function is causing the unit test errors. If you use `try` then you can do

```suggestion
    if (inherits(quantities_i, "try-error")) {
```

R/loo_moment_matching.R
```diff
@@ -286,14 +286,20 @@ loo_moment_match_i <- function(i,
     # 1. match means
     trans <- shift(x, uparsi, lwi)
     # gather updated quantities
-    quantities_i <- update_quantities_i(x, trans$upars,  i = i,
+    quantities_i <- tryCatch(update_quantities_i(x, trans$upars,  i = i,
```
jgabry, line 289:
```suggestion
    quantities_i <- try(update_quantities_i(x, trans$upars,  i = i,
```

Here you were using `tryCatch` but in the other cases below you use `try`. I would just use `try` everywhere for consistency (it's also simpler to use than `tryCatch`).

jgabry (changes requested):
This seems like a good idea. I made a few small suggestions in review comments.

R/loo_moment_matching.R
avehtari: Oops, `is_try_error` is from `brms` and I forgot to run tests in fresh R session

avehtari:
Thanks for fixes! I need to remember to run tests in *fresh* R session. I was learning `try` and `TryCatch` and for a moment thought I might need the latter, and for some reason didn't see that it was still there

jgabry (approved):

PR #264, opened 2024-03-28
Fix E_loo_khat error when posterior::pareto_khat returns NA

### LINKED ISSUE #263: `loo_R2` can fail on examples where it worked before
The newest loo package seems to be less robust than previous versions when calculating `loo_R2`. From the NES [example](https://avehtari.github.io/ROS-Examples/NES/nes_logistic.html), 
```
data("nes", package = "rosdata")
ok <- nes$year==1992 & !is.na(nes$rvote) & !is.na(nes$dvote) & (nes$rvote==1 | nes$dvote==1)
library(rstanarm)
fit_1 <- stan_glm(rvote ~ income, family=binomial(link="logit"), data=nes92, refresh = 0)
loo_R2(fit_1)
```
This fails with the warning message "Input contains infinite or NA values, Pareto smoothing not performed." and error message "Error: $ operator is invalid for atomic vectors". Neither of these messages is very helpful. The warning results from `posterior:::should_return_NA` but it is due to the input being within `tol` of a constant rather than infinite or NA values. The error comes from `loo:::.E_loo_khat_i` due to the fact that  `posterior::pareto_khat` returns the scalar `NA` rather than a list in this situation, so trying to select the `$khat` element fails.

This all worked with loo 2.6.0.

jgabry:
closes #263

This avoids the error mentioned by @bgoodri in #263 by first checking if `posterior::pareto_khat` returns a list or NA.

### DIFF
--- a/R/E_loo.R
+++ b/R/E_loo.R
@@ -290,12 +290,18 @@ E_loo_khat.matrix <- function(x, psis_object, log_ratios, ...) {
 .E_loo_khat_i <- function(x_i, log_ratios_i, tail_len_i) {
   h_theta <- x_i
   r_theta <- exp(log_ratios_i - max(log_ratios_i))
-  khat_r <- posterior::pareto_khat(r_theta, tail = "right", ndraws_tail = tail_len_i)$khat
+  khat_r <- posterior::pareto_khat(r_theta, tail = "right", ndraws_tail = tail_len_i)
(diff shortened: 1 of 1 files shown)


avehtari:
Doesn't look nice, but I don't see how else to do it

avehtari (approved):

PR #266, opened 2024-04-12
update psis ref + some minor typo fixes

avehtari:
updated psis ref with final journal info + link, plus fixed some minor typos


jgabry (approved):
Thanks! I just regenerated the documentation files and pushed them so the unit tests will run again. We can merge when they finish.

PR #268, opened 2024-04-26
fix equations in loo-glossary

avehtari:
At some point I had added \eqn{}, but forgot to change R  code to LaTeX equations, This fixes that


jgabry:
Thanks! Back from my leave, so finally checking in on the repos again.

PR #269, opened 2024-05-20
Fix issue with pareto_khat output no longer being a list

n-kall:
The current github version of posterior now returns an unnamed numeric from `pareto_khat`. This PR should add compatibility for this, with a check that if the output is a list, it will take the `khat` element instead to retain compatibility with older posterior versions.

ping @avehtari


avehtari:
Looks good

Issue #272, opened 2024-07-04
"Error in if (varx == 0) { : missing value where TRUE/FALSE needed" with fit$loo() but not loo::loo(fit$draws("log_lik"))


mhollanders:
Hi,

Unfortunately I don't have a reproducible example because this is just popping up with a few versions of a big model. When I fit the model with cmdstanr, the following works:

```
> loo::loo(fit_mvn$draws("log_lik"))

Computed from 400 by 590 log-likelihood matrix.

         Estimate     SE
elpd_loo -31300.8 3489.9
p_loo       323.5   14.9
looic     62601.5 6979.8
------
MCSE of elpd_loo is NA.
MCSE and ESS estimates assume independent draws (r_eff=1).

Pareto k diagnostic values:
                          Count Pct.    Min. ESS
(-Inf, 0.62]   (good)     367   62.2%   37      
   (0.62, 1]   (bad)      177   30.0%   <NA>    
    (1, Inf)   (very bad)  46    7.8%   <NA>    
See help('pareto-k-diagnostic') for details.
```

But this doesn't:

```
> fit_mvn$loo()
Error in if (varx == 0) { : missing value where TRUE/FALSE needed
```

I checked and there's no NAs anywhere:

```
> fit_mvn$draws("log_lik") |> anyNA()
[1] FALSE
```

Does anyone have any idea? Sorry I can't be more helpful with reproducible code. I'd be happy to share the Stan program with simulation file.

Thanks,

Matt

jgabry:
Thanks for reporting this, that's strange. I'm glad at least `loo::loo(fit_mvn$draws("log_lik"))` works so this doesn't prevent you from using loo. I'm not sure where that error is coming from. I don't think loo or cmdstanr has a line of code containing `if (varx == 0)` so I'm guessing this is coming from another package that's being used internally? Do you by any chance have the traceback() available so we can see more detail about the source of the error message?

mhollanders:
Hey, thanks for the help. traceback() gives me the following:

```
> fit$loo()
Error in if (varx == 0) { : missing value where TRUE/FALSE needed
> traceback()
8: posterior::autocovariance(sims[, i])
7: FUN(X[[i]], ...)
6: lapply(1:chains, FUN = function(i) posterior::autocovariance(sims[, 
       i]))
5: FUN(array(newX[, i], d.call, dn.call), ...)
4: apply(x, 3, ess_rfun)
[...]

PR #273, opened 2024-08-04
avoid under and overflows in stacking

avehtari:
Made the stacking target function and gradient to work with wider range of log-densities. With that high range, it's likely that the optimization is dominated by one or few observations and stacking is not that useful, but at least a hard to interpret error is avoided and the code should be faster, too. Related to #272. Passed tests.


jgabry (approved):
Looks good. I just pushed one tiny commit to remove some extra parentheses.

Issue #249, opened 2024-02-27
Use more functions from the posterior package


avehtari:
Release 2.7.0 imports posterior. We could use more posterior package functions. This would reduce possibility of having different implementations and thus reduce maintenance burden. Potential functions to replace
- `ess_rfun
    - `fft_next_good_size`
    - `autocovariance`
- part of `do_psis_i`
    - `psis_smooth_tail`
    - `n_pareto`
    - `enough_tail_samples`
- `ps_khat_threshold`
- `pareto_smooth`
    - `.pareto_smooth_tail`
    - `as_one_logical`
    - `should_return_NA`
    - `is_constant`
- `ess_tail`
    - `ess_quantile`
    - `.ess_quantile`
    - `.ess`
    - `.split_chains`
- `gpdfit`
    - `lx`
    - `adjust_k_wip`
    - `qgpd`

PR #289, opened 2025-06-12
Use testthat 3e and rely on posterior for ESS

VisruthSK:
One test was failing due to minor numerical differences in algorithms: average diff: 0.00975, largest diff: 0.01472. I updated the expected test result at reference-results/relative_eff.rds.

posterior::autovariance is already used.

Starts work on #249


VisruthSK:
Is there a neat way to update all expected outputs to the new values so that tests pass?

jgabry:
I think if we updated to using the newer `testthat::expect_snapshot()` then there's a way to update all of them at one. The older `expect_equal_to_reference()` doesn't seem to have that ability unfortunately. At some point we should update to snapshot testing (it wasn't part of testthat when I wrote these tests a long time ago). 

If you want to do that as part of this PR we could, but also fine to hold off on that and just update the reference files.

VisruthSK:
I think swapping to snapshots is worthwhile, so I'll get started on that. It'll make this PR easier so might as well include it here. I briefly tried updating the reference files but I wasn't able to get the reference files to update easily.

jgabry:
Thanks! Is this ready for me to take another look or are you still working on this? Either way is fine, just checking.

jgabry:
I think this looks good. Will merge now!

PR #291, opened 2025-06-12
Format project using Air

VisruthSK:
Use Posit's new opinionated formatter, [Air](https://posit-dev.github.io/air/formatter.html) to [review pull requests](https://posit-dev.github.io/air/integration-github-actions.html#format-with-github-suggestions) to ensure they conform to the style as well as format the entire project once.

What do you think? @jgabry @avehtari

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -25,3 +25,5 @@ vignettes/loo2-non-factorized_cache/*
 ^_pkgdown\.yml$
 ^pkgdown$
 ^touchstone$
+^[.]?air[.]toml$
+^\.vscode$
--- a/.github/workflows/format-suggest.yaml
+++ b/.github/workflows/format-suggest.yaml
@@ -1,0 +1,46 @@
+# Workflow derived from https://github.com/posit-dev/setup-air/tree/main/examples
+
+on:
+  # Using `pull_request_target` over `pull_request` for elevated `GITHUB_TOKEN`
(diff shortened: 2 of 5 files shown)


jgabry:
(Oops accidentally closed the PR, I meant to leave it open to continue the discussion)

VisruthSK:
I think formatting is simple enough to add/maintain and low-risk enough that breaking changes don't matter as much--worst case one could just remove the action and there's no cost.

I think the action reformats the entire project `run: air format . --check`. It's rather fast; when I ran it on `loo` the whole thing was done in under a second or two. So I think if the formatting rules changed, new PRs would reformat the entire project anyway, so the formatting would be consistent throughout.

VisruthSK:
Ah that makes sense, thank you for clarifying. I didn't think of the diff clutter, but now that you mention it, that does seem quite distracting and not worth it for now.

Do you think it'd make sense to push off formatting PRs till Air is stable, but format the entire project on every release? This would have the problem that you mentioned of odd formatting in between releases though. Or do you think it makes more sense to just wait till Air is stable before using it at all?

jgabry:
I would lean towards waiting until it's stable to use it, but I'll think more about it. And maybe they will do a 1.0 release soon. We can definitely leave this PR open in the meantime. Or maybe convert it to a draft PR if it's possible to convert after opening it (or close and reopen as a draft). I definitely think this would be a good thing to do when it's ready.

Issue #299, opened 2025-08-31
Add normal approximation based probabilities to loo_compare


avehtari:
Paper [Uncertainty in Bayesian Leave-One-Out Cross-Validation Based Model Comparison](https://arxiv.org/abs/2008.10296) describes when `elpd_diff` and `se_diff` based normal approximation can be expected to be well calibrated. The paper includes case studies with the code at https://users.aalto.fi/~ave/casestudies/LOO_uncertainty/loo_uncertainty.html. In that code, the probability that a model has worse elpd_loo than the best model has been computed separately and the text includes comments on when we can expect the normal approximation to be well calibrated. We could add the probability that a model has worse elpd_loo and diagnostic message to loo_compare() output

Example outputs would look like

Small data:
```
> loo_compare(loo1, loo2, loo3, loo4)
    elpd_diff se_diff p_worse diag_pnorm
M_4       0.0     0.0      NA    N < 100
M_3      -3.9     1.6    0.99    N < 100
M_1      -4.0     2.4    0.95    N < 100
M_2      -4.7     2.5    0.97    N < 100
```

Outliers (khat threshold depends on the number of observations):
```
> loo_compare(M_1, M_2, M_3)
    elpd_diff se_diff p_worse       diag_pnorm
M_3       0.0     0.0      NA                 
M_2     -13.5     9.8    0.92 khat_diff > 0.54
M_1     -78.1    20.8    1.00 khat_diff > 0.54
```

All good:
```
loo_compare(M_1t, M_2t, M_3t)
     elpd_diff se_diff p_worse diag_pnorm
M_3t       0.0     0.0      NA           
M_2t     -44.8     8.6    1.00           
M_1t    -118.7    15.9    1.00           
```

Models that have very similar predictions
```
> loo_compare(loo(M_3), loo(M_4))
    elpd_diff se_diff p_worse          diag_pnorm
M_4       0.0     0.0      NA                    
M_3      -2.2     3.2    0.76 similar predictions
```

I have a branch that clarifies the diagnostics, and illustrates some things probably need more thought. Now I added `diag_pnorm` as string, which meant I had to switch to use data.frame, and I want to show one digit for `elpd_diff` and `se_diff` but two digits for `p_worse`

florence-bockting:
Fixed by #300

PR #300, opened 2025-08-31
add diff diagnostics

avehtari:
## Description
Fixes issue [#299](https://github.com/stan-dev/loo/issues/299)

In this PR we are making changes to the output structure returned by `loo_compare()`. The main changes are a transition from a matrix to a data frame and the addition of several new columns. Here is an example of the updated output structure:

```
> loo_compare(w1, w2)
  model elpd_diff se_diff p_worse diag_diff diag_elpd
 model1       0.0     0.0      NA                    
 model2      -4.1     0.1    1.00   N < 100          
 
Diagnostic flags present.
See ?`loo-glossary` (sections `diag_diff` and `diag_elpd`)
or https://mc-stan.org/loo/reference/loo-glossary.html.
```

More context on the motivation behind these changes can be found in the updated case study ["Uncertainty in Bayesian LOO-CV Model Comparison"](https://users.aalto.fi/~ave/casestudies/LOO_uncertainty/loo_uncertainty.html).

## TODOs
+ [x] implement changes and corresponding tests
+ [x] update NEWS.md
+ [x] fix all reverse dependency issues (see [Comment](https://github.com/stan-dev/loo/pull/300#issuecomment-4276271699))

## Comment
We will merge this PR such that it is available in the main branch even though there is still one open PR with respect to the reverse dependency checks (see [Comment](https://github.com/stan-dev/loo/pull/300#issuecomment-4276271699)). We tried to contact the maintainer multiple times but so far without success.


R/loo_compare.R
```diff
@@ -116,7 +116,30 @@ loo_compare.default <- function(x, ...) {
   diffs <- mapply(FUN = elpd_diffs, loos[ord[1]], loos[ord])
   elpd_diff <- apply(diffs, 2, sum)
   se_diff <- apply(diffs, 2, se_elpd_diff)
-  comp <- cbind(elpd_diff = elpd_diff, se_diff = se_diff, comp)
+  # compute probabilities that a model has worse elpd than the best model
+  # using a normal approximation (Sivula et al., 2025)
+  p_worse <- pnorm(0, elpd_diff, se_diff)
```
jgabry:
```suggestion
  p_worse <- stats::pnorm(0, elpd_diff, se_diff)
```

will avoid an R cmd check note or warning

R/loo_compare.R
```diff
@@ -131,9 +154,10 @@ loo_compare.default <- function(x, ...) {
 #' @param digits For the print method only, the number of digits to use when
 #'   printing.
 #' @param simplify For the print method only, should only the essential columns
-#'   of the summary matrix be printed? The entire matrix is always returned, but
-#'   by default only the most important columns are printed.
-print.compare.loo <- function(x, ..., digits = 1, simplify = TRUE) {
+#'   of the summary matrix be printed? The entire matrix is always returned, bu#' @param pnorm For the print method only, should we include the normal
+#'   approximation based probability of model having worse performance than
+#'   the best model
```
jgabry: The end of doc for the `simplify` argument got deleted and `#' @param pnorm` should be on a new line. 

```suggestion
#'   of the summary matrix be printed? The entire matrix is always returned, but
#'   by default only the most important columns are printed.
#' @param pnorm For the print method only, should we include the normal
#'   approximation based probability of each model having worse performance than
#'   the best model?
```

R/loo_compare.R
```diff
@@ -143,12 +167,16 @@ print.compare.loo <- function(x, ..., digits = 1, simplify = TRUE) {
   } else if (NCOL(xcopy) >= 2 && simplify) {
      xcopy <- xcopy[, c("elpd_diff", "se_diff")]
   }
-  print(.fr(xcopy, digits), quote = FALSE)
-  invisible(x)
+  if (p_worse) {
```
jgabry: This will cause an error when printing because the argument name is `pnorm`. But maybe the argument name should be `p_worse`? Either way, either the argument name needs to change or this `if` clause needs to change.

jgabry (commented):
* I think this is a good idea. It's possible that this doesn't actually break any packages that depend on loo. I can check by running reverse dependency checks once the errors in the code are fixed (see individual comments). 

* The tests need to be updated, but we can wait to do that until we check that it doesn't break any reverse dependencies 

* We should put some information about this in main `loo_compare` documentation, not only the print method

* I actually think it would be good to print these probabilities by default. What was the reason you turned it off by default? (Actually right now the printing will error by default, but once fixed it will be off by default). I don't think the printing should affect backwards compatibility (or am I missing something?). It would be adding more columns could potentially affect something, right? Or was there a different reason you turned off printing the probabilities by default?

jgabry:
Setting `simplify=FALSE` when calling the print method also errors due to a weird interaction with the `p_worse`. But I think we can get rid of the `simplify` argument. I've never seen anyone use it and I doubt it will break any packages (we can find out by reverse dependency checks). I will push a commit fixing this.

R/loo_compare.R
```diff
@@ -116,7 +122,34 @@ loo_compare.default <- function(x, ...) {
...
+  # diagnostics to assess whether the normal approximation can be trusted
+  N <- nrow(diffs)
+  if (N < 100) {
+    # small N (Sivula et al., 2025)
+    diag_pnorm <- rep("N < 100", length(elpd_diff))
+    diag_pnorm[elpd_diff == 0] <- ""
+  } else {
+    diag_pnorm <- rep("", length(elpd_diff))
+    # similar predictions (Sivula et al., 2025)
+    diag_pnorm[elpd_diff > -4 & elpd_diff != 0] <- "similar predictions"
```
jgabry: Should this be `|elpd_diff| < 4` instead of "similar predictions" for consistency with the others? Or you could change the others to "small data" and "possible outliers". But right now there's a mix.

jgabry:
I fixed the tests so this is passing now. I will run reverse dependency checks over the weekend hopefully. I just put one question for you @avehtari in the review comments about how to label to diagnostic.

avehtari:
> I like diagnostic the best, but do you think if we name it diagnostic (as opposed to diag_p_worse) then users will think it also means elpd_diff can't be trusted?

1. `elpd_diff` should be interpreted together with `diff_se` as otherwise it's not clear how big `elpd_diff` is big or how big the error might
2. in case of high `khat_diff`, `elpd_diff` is very high

Considering above points, I feel like `diag_p_worse` focuses too much on the probability and not all `diff` columns. As the returned data frame has also `elpd_loo`, I would now think `diag_dff` would be the best choice

>  I reformatted the documentation a little bit so that now these diagnostics are in a bulleted list that's easier to read (as opposed to paragraph form) and the user can quickly check what the diagnostic means:

Looks better now!

> loo_subsample has its own loo_compare method and print method. I guess there were enough differences that it made sense for @MansMeg to implement a separate method.

The columns and the print function should look the same, but the loo_compare includes subsampling estimator computation for se, so it is different. I guess it could be possible to combine them. We can add the same columns, but I need to think a bit about the p_worse and Pareto-k diagnostic in case of subsampling

jgabry:
> Considering above points, I feel like `diag_p_worse` focuses too much on the probability and not all `diff` columns. As the returned data frame has also `elpd_loo`, I would now think `diag_dff` would be the best choice

I like `diag_diff`. 

 > The columns and the print function should look the same, but the loo_compare includes subsampling estimator computation for se, so it is different. I guess it could be possible to combine them. We can add the same columns, but I need to think a bit about the p_worse and Pareto-k diagnostic in case of subsampling

Ok for now I think we can just make sure that both return a data frame and have a `model` column (if we add that). We can then add the `p_worse` and `diag_diff` columns at some point once you've thought more about it. 

Regarding switching from row names to a `model` column, I think the extra packages that are broken are not too hard to fix. I think it's about 3 or 4 additional packages that beyond the ones that are already affected by the current PR. Do you want me to merge my branch into this one?

jgabry:
Here's an example of how rstanarm output looks using this branch. 

There are two methods. The first just uses regular `loo_compare()` and so it's no different than just using the loo package: 

```r
fit_wt <- stan_glm(mpg ~ wt, data = mtcars, refresh = 0)
fit_wt_cyl <- stan_glm(mpg ~ wt + cyl, data = mtcars, refresh = 0)
fit_wt_cyl_gear <- stan_glm(mpg ~ wt + cyl + gear, data = mtcars, refresh = 0)

loo_wt <- loo(fit_wt)
loo_wt_cyl <- loo(fit_wt_cyl)
loo_wt_cyl_gear <- loo(fit_wt_cyl_gear)
loo_compare(loo_wt, loo_wt_cyl, loo_wt_cyl_gear)
```

```
           model elpd_diff se_diff p_worse diag_diff
      fit_wt_cyl       0.0     0.0      NA          
 fit_wt_cyl_gear      -1.0     0.8    0.90   N < 100
          fit_wt      -5.1     2.8    0.97   N < 100
```

The second method adds `loo` objects to the fitted model objects and then puts them in a list: 

```r
fit_wt$loo <- loo_wt
fit_wt_cyl$loo <- loo_wt_cyl
fit_wt_cyl_gear$loo <- loo_wt_cyl_gear

loo_compare(stanreg_list(fit_wt, fit_wt_cyl, fit_wt_cyl_gear), detail = TRUE)
```

```
Model formulas: 
 fit_wt:  mpg ~ wt
 fit_wt_cyl:  mpg ~ wt + cyl
 fit_wt_cyl_gear:  mpg ~ wt + cyl + gear

Model comparison based on LOO-CV: 
           model elpd_diff se_diff p_worse diag_diff
      fit_wt_cyl       0.0     0.0      NA          
 fit_wt_cyl_gear      -1.0     0.8    0.90   N < 100
          fit_wt      -5.1     2.8    0.97   N < 100
```

avehtari:
I have tested model as column name, and it works well

I added yet another column `diag_elpd` to show if there were high Pareto-k's in PSIS-LOO computation. We had discussed this with @n-kall some time ago. Sometimes people look at the `loo_compare()` output without checking whether there were high khats. You can see how this looks like and how to interpret the results are illustrated in https://users.aalto.fi/~ave/modelselection/roaches_diag.html

The roaches example has an interesting case where in the end two models are compared, and the khat for pointwise elpds for both models is <0.5, but for the khat>0.5 for the pointwise differences. This is possible due to dependency from dependency of both models predicting the same target. This however goes slightly beyond what is said in the LOO uncertainty paper.

You can see how the comparison tables look like with tinytable at
https://users.aalto.fi/~ave/casestudies/LOO_uncertainty/loo_uncertainty_diag.html

R/loo_compare.R
```diff
@@ -101,54 +126,122 @@ loo_compare.default <- function(x, ...) {
...
+    khat_diff[elpd_diff != 0] <- apply(
+      diffs[, elpd_diff != 0, drop = FALSE], 2,
+      function(x) ifelse(length(unique(x)) <= 20, NA, posterior::pareto_khat(x, tail = "both")
+    ))
+    diag_diff[khat_diff > 0.5] <- "khat_diff > 0.5"
+  }
+
+  # get khats for PSIS
+  khat_psis <- sapply(loos[ord],
+                      \(loo) {
```
jgabry:
```suggestion
                      function(loo) {
```

This syntax is only available in more recent versions of R (I think >= 4.1), but loo can work with older R versions so we shouldn't use the new syntax unless we want to restrict loo to R>=4.1. If we want to do that we need to increase the R version listed in the DESCRIPTION file.

R/loo_compare.R
```diff
@@ -41,9 +38,41 @@
...
+#'   likely to be too large (small data or similar predictions) or too small
+#'   (outliers). `elpd_diff` and `se_diff` are still indicative of the
+#'   differences and uncertainties, and for example, if `|elpd_diff|` is
+#'   many times larger than `se_diff` the difference is quite certain.
+#'   While `khat > 0.5` indicates possibility of outliers, it is also
+#'   possible that both models compared seem to be well specified based
+#'   on model checking, but the pointwise ELPD differences have such thick
+#'   tails that the normal approximation for the sum is not good (Vehtari et
+#'   al., 2024). Note that `khat`s in `diag_diff` column are different from
+#'   the `khat`s in PSIS-LOO  diagnostic.
```
jgabry: I don't think it's explained anywhere how `khat` in `drag_diff` is computed, so users may be confused about having two different `khat`s,

avehtari:
> You can see how the comparison tables look like with tinytable at https://users.aalto.fi/~ave/casestudies/LOO_uncertainty/loo_uncertainty_diag.html
>
> It looks good, although none of the examples in this have anything in the diag_elpd column, so it doesn't show what that looks like.

That's why I linked also the roaches case study, which has stuff in diag_elpd

avehtari:
Maybe change `khat > 0.5` to `k_diff > 0.5` and `khat_psis > 0.7` to `k_psis > 0.7` to make them more different and also shorter. Also the current print.loo says "Pareto k diagnostic values", that is, there is not hat

jgabry:
> Maybe change `khat > 0.5` to `k_diff > 0.5` and `khat_psis > 0.7` to `k_psis > 0.7` to make them more different and also shorter. Also the current print.loo says "Pareto k diagnostic values", that is, there is not hat

Ok yeah that sounds good.

jgabry:
I made some small doc edits, fixed the tests, and moved the computation of `diag_diff` and `diag_elpd` to separate internal functions (defined at the end of the same R file). This way `loo_compare()` itself is a little less messy.

avehtari:
We discussed this PR in our group meeting and I think at this point we should drop `k_diff` diagnostic as 1) it may produce too many false positives with typical data sizes, and 2) it's not directly connected to the LOO uncertainty paper point about model misspecification. Although it is likely to be useful, we need more experiments to be able to provide better advise how to interpret it and what additional checks can be done. We can keep the other diagnostics at this time.

avehtari:
I have updated a case study using this PR https://users.aalto.fi/~ave/casestudies/LOO_uncertainty/loo_uncertainty_diag.html

There are now more explanations on how to interpret the diagnostics

avehtari (commented):
Excellent loo-glossary text. Just one comment about Bengio and Grandvalet reference

R/loo_compare.R
```diff
@@ -101,54 +121,94 @@ loo_compare.default <- function(x, ...) {
...
+  print(x2, quote = FALSE, row.names = FALSE)
+
+  # show glossary for diagnostic flags
+  has_diag <- any(nzchar(x[["diag_diff"]], keepNA = FALSE), na.rm = TRUE) ||
+              any(nzchar(x[["diag_elpd"]], keepNA = FALSE), na.rm = TRUE)
+  if (has_diag && p_worse) {
+    message(
+      "\nDiagnostic flags present. ",
+      "See ?`loo-glossary` (sections `diag_diff` and `diag_elpd`) ",
+      "or https://mc-stan.org/loo/reference/loo-glossary.html."
```
avehtari: This message is now 140 chars long, and should be divided to several lines, not exceeding 80 chars per line

florence-bockting: thanks for pointing this out. Fixed.

jgabry:
@florence-bockting @avehtari Thank you in advance for doing the new round of reverse dependency checks. 
@florence-bockting please also officially add yourself to the DESCRIPTION file as part of this PR! Or separately from this PR, whatever you prefer.

florence-bockting:
Re-running reverse dependency checks. In the following a list of packages where issues occur:

- [x] BAMBI 2.3.6
   + extracted model names as rownames which changed now due to new "model" column
   + Public PR: [PR#5](https://github.com/c7rishi/BAMBI/pull/5)
- [x] BayesERtools 0.2.4
   + extracted model names as rownames which changed now due to new "model" column
   + Public PR: [PR#20](https://github.com/Genentech/BayesERtools/pull/20)
- [x] clmstan 0.1.1
   + checked in tests for `is.matrix`; changed to `is.data.frame` in case "model" column exists
   + Public PR: [PR#1](https://github.com/t-momozaki/clmstan/pull/1) [22/04/2026]
   + First E-mail reminder: 29/04/2026; Second E-mail reminder: 08/05/2026
- [x] disbayes 1.1.0
   + uses in test rownames for extracting model; changed to use of column value if column "model" exists
   + Public PR: [PR#3](https://github.com/chjackson/disbayes/pull/3)
- [x] flocker 1.0.0
   + test for rownames; tutorial shows output of loo_compare as comment -> changed corresponding test and output comment
   +  Public PR: [PR#116](https://github.com/jsocolar/flocker/pull/116)
- [x] ubms 1.2.8
   + assumes a type matrix for extracting models; update to switch between data.frame and matrix depending whether "model" column exists
   +  Public PR: [PR#88](https://github.com/ecoverseR/ubms/pull/88)
   
@avehtari: I have now worked out fixes for all affected libraries which are currently still local. When I understand the process correctly, then we can now merge this PR into master and I will make the PRs in the respective packages public. Or some other thoughts on this?

florence-bockting:
@VisruthSK The touchstone build process fails in this repo. Can you please have a look at this? Thank you.

PR #305, opened 2025-10-17
use posterior::gpdfit and posterior::qgeneralized_pareto()

avehtari:
Since we do use `posterior` package elsewhere, 
replace gpdfit() and qgpd() with posterior::gpdfit and posterior::qgeneralized_pareto()

This will make maintenance easier. For example, there is a PR for posterior::gpdfit() to make it more robust and posterior::qgeneralized_pareto() already is more robust than qgpd()

This is related to #249, but timing is due to the fix PR for posterior::gpdfit()


jgabry (commented):
I think the R cmd check error is because the NAMESPACE file is still trying to export gpdfit from loo. If you run devtools::document() it should fix this and then we can see if all the tests pass.

avehtari:
At least in github only I and Juho Timonen have used loo::gpdfit, so I guess it can be removed

jgabry:
The error now is because apparently these functions aren't exported from posterior, they're just internal functions. That would be easy to change but it would require doing a CRAN release of posterior.

jgabry:
Ok yeah good call

jgabry:
Aside from needing to reexport and add the tests back, this looks good.

jgabry (approved):
Looks good, thanks. Although I wish we had #348 set up to make sure the posterior versions aren't less efficient. For example, maybe they add overhead due to doing more type and argument checking that we skipped in the previous loo implementation?

VisruthSK:
Hm fair enough I guess we could just run the toplevel function. I suppose that's what I've been doing too--but it would be nice to get a little but of a profile by breaking down the call into some cogent sub-calls--i.e., if there's a PR which speeds up PSIS, it'd be nice to see that most things don't change but PSIS takes half the time or whatever.

VisruthSK:
I guess we could shelve the finer touchstone tests for now and revist once we already have working pipeline--writing in #352.

jgabry:
> This is how benchmark results would change (along with a 95% confidence interval in relative change) if [cddb7cc](https://github.com/stan-dev/loo/commit/cddb7cc45c0245be3e1acf361a7cf194498852b0) is merged into master:
> 
> * 🚀loo_function: 1.99s -> 1.74s [-12.9%, -12.07%]
> * 🚀loo_matrix: 1.53s -> 1.28s [-17.25%, -16.27%]

Ok so using the posterior versions makes it faster?

VisruthSK:
Yup looks like it. We should probably increase the number of repetitions if we care about the actual reduction amount. Kinda odd to have confidence intervals in a Stan project...

jgabry:
I agree we should increase the repetitions. We’ll forgive the author of the touchstone package for the lack of a Bayesian analysis ;)

VisruthSK:
New package, `touchstan`?

Do you want to bump up the repeats and see how the numbers shake out, or do you think this is good to merge?

avehtari:
Confidence interval is acceptable, but definitely there is not enough accuracy to justify more than two significant digits
> 1.53s -> 1.28s [-17.25%, -16.27%]

jgabry:
> Do you want to bump up the repeats and see how the numbers shake out, or do you think this is good to merge?

I would increase the repetitions if the code is pretty fast. But it's still taking ~1 hr (https://github.com/stan-dev/loo/actions/runs/24422791263/job/71349231024?pr=305). Is that because it's not caching the packages?

VisruthSK:
Yes I think it's building a lot of packages from source. I think I know a way around that if reruns are slow. If it's back to the times we saw before, about 5 minutes for a whole run at this repetition count, what do you think would be a good number of reps? How slow is okay, 30 min?

jgabry:
Yeah, I think 20-30 minutes is fine

VisruthSK:
@jgabry good to merge? This is with n=60 reps per branch. Slow touchstone action because I moved to new ubuntu and R version.

jgabry:
Thanks, yeah I think this one is good to go. Merging now.

PR #307, opened 2025-10-19
Print unequal sample sizes in loo_compare

### LINKED ISSUE #296: Include more information in error about different number of data points
Suggested by @bgoodri: when `loo_compare()` errors due to models have different numbers of data points, it would be helpful to the user if we print the number of data points instead of just saying they're different. 

https://github.com/stan-dev/loo/blob/734c3be9f9a34859938a51324d8c837a44ea14d9/R/loo_compare.R#L191

VisruthSK:
Fixes #296.


jgabry (commented):
Thanks! This looks good at first glance but I haven't had a chance to try it out yet. Busy day for me today but will try it out as soon as I can.

PR #328, opened 2026-02-21
Clarify print output to specify posterior draws and log-likelihood terms

### LINKED ISSUE #198: Specify Sample Size a Bit More Clearly
The current phrasing contains the same information but is a bit harder to interpret:
```
Computed from 4000 by 262 log-likelihood matrix
```
ParetoSmooth.jl is more explicit about this, with a phrasing that mentions "n observations and m posterior samples." I've found this prevents mistakes -- it's very common for users to accidentally misspecify what it is they want to leave out, e.g. by placing all their observations in one multivariate normal, which causes confusion when the whole dataset is left out. I think this might be a more useful message.

ishaan-arora-1:
Hey! This addresses #198, where the current print output like

```
Computed from 4000 by 262 log-likelihood matrix
```

was flagged as ambiguous — it's not immediately clear which dimension is the number of posterior draws and which is the number of observations/log-likelihood terms. As mentioned in the issue, ParetoSmooth.jl already uses a more explicit format, and this kind of ambiguity can hide bugs (e.g. a user accidentally placing all observations in one multivariate normal).

## Changes

The new output format is:

```
Computed from 4000 posterior draws and 262 log-likelihood terms.
```

This is close to what @avehtari proposed in the issue comments (`"Computed from 4000 posterior draws by 262 log-likelihood terms matrix"`), but I dropped "matrix" since the user doesn't really need to know it's a matrix — they care about what the numbers mean.

For log-weight objects (raw `psis`/`tis`/`sis`):
```
Computed from 1000 posterior draws and 32 log-weight terms.
```

For subsampled loo:
```
Computed from 4000 posterior draws and 100 subsampled log-likelihood
terms from 3020 total observations.
```

### Updated methods

- `print_dims.psis_loo`
- `print_dims.importance_sampling`
- `print_dims.importance_sampling_loo`
- `print_dims.waic`
- `print_dims.psis_loo_ss`
- `print_dims.elpd_generic`

### Also updated

- All corresponding tests in `test_print_plot.R` and `test_loo_subsampling_cases.R`
- Snapshot files for `psis.md`, `tisis.md`, and `loo_moment_matching.md`
- Vignette output in `loo2-with-rstan.Rmd` and `loo2-large-data.Rmd`
- `NEWS.md` entry

Fixes #198


ishaan-arora-1:
> Yes, do you want to take a stab at it?
> 
> The [tests](https://github.com/stan-dev/loo/actions/runs/22314445496/job/64557780403?pr=328#step:6:164) are expecting "subsampled log-likelihood values" but the new message is "subsampled log-likelihood terms", so the test case should reflect that.

sure, would love to :)

VisruthSK (approved):

VisruthSK:
Lgtm. If Aki likes the language I think this is good to merge.

The only thing I can think of is maybe the tests that you just had to fix should be snapshot tests instead, but that change needn't be made in this PR.

florence-bockting:
Hi, just a quick check-in on the status of this PR. I think @avehtari had some further [comments on the information text](https://github.com/stan-dev/loo/pull/328#issuecomment-4083737240) that still need to be addressed. Do you still have time to work on this, @ishaan-arora-1, or would you like us to take over?

florence-bockting:
> The text for subsampling case in the latest commit is too long. For example, with the example in tests, the text is 225 characters. 1) With `loo_subsample()` I think that both the fast and slow part use the same number of posterior draws, so that part would not need to be repeated. 2) If the more compact text is still over 80 characters, it would be better to divide in two lines (hopefully three lines is not needed).
> 
> Before making any additional commits, make a proposal for new text here in the discussion, and after we have agreed on good new text, then you can make a commit for that. This will save you time, as you don't need to change the tests before we have agreed on the text.

Hi @avehtari, what do you think about the following user message:

```
cat(
  "Subsampling loo uses 1) faster, more biased computation with",
  length(x$loo_subsampling$elpd_loo_approx), "terms,\n",
  "and 2) slower, more accurate computation with",
  nobs(x), "subsampled terms.\n"
)
```

florence-bockting:
> Hi, just a quick check-in on the status of this PR. I think @avehtari had some further [comments on the information text](https://github.com/stan-dev/loo/pull/328#issuecomment-4083737240) that still need to be addressed. Do you still have time to work on this, @ishaan-arora-1, or would you like us to take over?

Update: I will take over at this point as we got no feedback from the contributor.

avehtari:
> Hi @avehtari, what do you think about the following user message:

OK for me

PR #340, opened 2026-03-22
Export srs_diff_est()

### LINKED ISSUE #333: export srs_diff_est
`loo_subsample()` can sometimes be over-complex. `srs_diff_est()` is a non-exported internal function used by `loo_subsample`. `srs_diff_est()` implements difference estimator Equations 7-9 of [Magnusson et al., 2020](https://proceedings.mlr.press/v108/magnusson20a.html). It would be useful to export this function and provide user targeted function with documentation.

vinniott:
Fixes #333

Note:
- There is no `@example` yet because I did not have time yet to get fully familiar with the whole package.
- I updated NEWS.md as suggested in CONTRIBUTING.md but I am not sure whether I did that correctly.


vinniott:
Interesting, do I understand it correctly that the `log_lik_matrix.rda` would be stored in loo/data ?
I will try to continue trying to implement this example towards the end of next week.

avehtari:
Instead of generating log_lik on the fly, we could use the stored wine log_lik which was added in https://github.com/stan-dev/loo/pull/352. This would make the example to run much faster. Need to check whether wine log_lik was added only for touchstone and was it too big for CRAN, @jgabry , @VisruthSK

vinniott:
The tests are failing so I will give a short summary of what I did.
First, I fetched all upstream updates and added them to this export-srs-diff-est feature branch of the PR.

Second, I renamed `generate-example_loglik-array.R` to generate-example_loglik-objects.R and also
`example_log_lik_array.R` to `example_log_lik_objects.R.`
Here, I added the `.example_wine_loglik_matrix` to `sysdata.rda` as you suggested above @jgabry 

Third, I added the example by @avehtari to `srs_diff_est()`.

Fourth, I did `devtools::document()`, `devtools::install()`, and when I then ran the example of `?srs_diff_est()`
it worked on my machine (e.g., screenshot from my RStudio):
Screenshot 2026-04-26 at 14 38 40

So I hope that I shouldnt be too far from getting the tests to pass. Do you have an idea why they might fail?

> Instead of generating log_lik on the fly, we could use the stored wine log_lik which was added in https://github.com/stan-dev/loo/pull/352.

Of course also an option! I just wanted give the `sysdata.rda` implementation a try as I was close to finishing it when you pointed this out. Though, the tests currently do not pass, of course.

VisruthSK:
> Instead of generating log_lik on the fly, we could use the stored wine log_lik which was added in #352. This would make the example to run much faster. Need to check whether wine log_lik was added only for touchstone and was it too big for CRAN, @jgabry , @VisruthSK

Yes to both. Clocks in at about 40MB, and is currently only in the touchstone directory.

VisruthSK:
Hi @vinniott! Thanks for working on this.

> So I hope that I shouldnt be too far from getting the tests to pass. Do you have an idea why they might fail?

I think we should figure out how we're storing/using the wine data before we fix the tests.

Apologize if you already know this, but if you click on the failed runs, you'll be able to scroll down till you see [R CMD check](https://r-pkgs.org/whole-game.html#check) output. Locally, you can run `devtools::check()` which will take a long time, but will approximate the same tests that are being run here. [Here's a link](https://github.com/stan-dev/loo/actions/runs/24956499095/job/73075585980?pr=340#step:6:205) to the main issue in this PR right now--I don't think the wine data is available for the package, so the example is failing when trying to find it. You can see that again further down, L216, when R CMD check is complaining about not finding a binding for `.example_wine_loglik_matrix`.

> Fourth, I did devtools::document(), devtools::install(), and when I then ran the example of ?srs_diff_est() it worked on my machine

I think the problem here might be that you ran some code locally before that to make `.example_wine_loglik_matrix` populate in your local environment. There's a way to restart R sessions in RStudio I think, and in the future if you run into a bug like this where things are okay locally but failing for someone else/on a runner, it might help if you try running the tests in a fresh R session.

Hope that slightly clears up why the tests are failing, and how to (hopefully) reproduce those failures locally.

vinniott:
HI @VisruthSK,
thank you so much for the elaborate help, really!
Yes, let's first clarify, how to store the wine loglik matrix.

If it's currently too big for CRAN as is, does this mean that neither touchstone nor sysdata.rda (as suggested above by @jgabry) are possible and that an alternative (data set or data storage method) is needed?

avehtari:
- There seems to be some mistake in this PR as it shows in "Files changed" and in commits many things that should not be part of this PR which makes reviewing it harder
- I think we should just show the code for how to use srs_diff_est in the doc, but not actually run the code and then we don't need to worry about how much generating log_lik takes or how much space it takes

vinniott:
Okay. There are now a lot of confusing commits. Here is what I did:

As suggested by @avehtari I have now simply added a commented out example.
All this happend in a clean, new feature branch with cherry-picked previous commits (https://github.com/stan-dev/loo/compare/master...vinniott:loo:clean-export-srs-diff-est).
I then `git reset --hard clean-export-srs-diff-est` onto this existing branch called `export-srs-diff-est`
because ChatGPT I recommended this over creating a new PR with the new clean branch.

I have two ettiquette questions:

- Do you generally recommend the `git reset` procedure over a new PR? Even in this case?

- What do I best do if I work on a feature/PR branch and then the upstream master updates? Should I merge those commits into my own master as well as my own feature branches? I am asking because this caused me a lot of headache here (though I learned a lot!). Maybe I did it in a wrong way?

VisruthSK:
@vinniott Thanks for making that new, clean branch! I just moved this PR to use that branch, sorry for not asking or anything but I hope you don't mind.

> Do you generally recommend the git reset procedure over a new PR? Even in this case?

Can't speak in general, but personally I didn't want to lose any discussion from this PR. Making a new, clean PR and linking to this one is also okay I guess, but since cleaning this PR up wasn't too hard (with the help of a LLM), I think updating a PR is cleaner.

> What do I best do if I work on a feature/PR branch and then the upstream master updates? Should I merge those commits into my own master as well as my own feature branches?

This is one of the reasons I don't like working on forks :) I think usually you should reabse feature branches to origin master, and keep fork's master clean and up to date with origin. You can do a merge commit too, but rebasing is cleaner IMO.

I glanced through [this forum post](https://github.com/orgs/community/discussions/153608#discussioncomment-12453674), and it seems good. I would also suggest using the GitHub UI if possible, then the `gh` CLI, then plain old `git`. GitHub Desktop might have some features for this too.

Hope this helps!

VisruthSK:
Separately, I think the example should be in a dontrun block instead of commented out.

vinniott:
Again, thank you @VisruthSK for the elaborate help!!

I have now implemented your suggestion of \dontrun{}.
This is currently in the newer, clean feature branch (https://github.com/stan-dev/loo/compare/master...vinniott:loo:clean-export-srs-diff-est)
Next, I will learn how to force-push it myself to this PR so you wont have to do it again.

I think then there is one last remaining question:
How should I rewrite the following line that @avehtari proposed:
`wine <- read.delim(root("winequality-red", "winequality-red.csv"), sep = ";") |>
  distinct()`
?

Do I understand it correctly, that `data/winequality.rda` could work so that the line in question becomes:
`wine <- data(winequality, package = "loo")`?
Or would you still recommend a procedure as with the log_lik_matrix above via the data-raw folder, @jgabry ?

VisruthSK:
>  Next, I will learn how to force-push it myself to this PR so you wont have to do it again.

I think you can just work off the export-srs-diff-est branch now, it's been cleaned up so the "clean" branch is redundant. You can just make your changes on export-srs-diff-est!

> How should I rewrite the following line that @avehtari proposed:...

Hm how about something like this in the example: 

```r
wine_scaled <- read.delim(
  "https://archive.ics.uci.edu/ml/machine-learning-databases/wine-quality/winequality-red.csv",
  sep = ";"
) |>
  unique() |>
  scale() |>
  as.data.frame()
```

That way we don't need to bundle the raw data nor the loglik in the package. With above code we can also drop `library(dplyr)` then I think, not that it matters.

vinniott:
Again, great input, @VisruthSK, thanks!
With that I think we now have a fully functional example and this PR might be ready for merge?

VisruthSK:
@vinniott I think you need to fetch and pull so that your branch is current and correct. I just requested a review from Aki, he might want some changes.

R/loo_subsample.R
```diff
@@ -1166,12 +1166,88 @@ loo_subsample_estimation_diff_srs <- function(x) {
...
+#' Modeling wine preferences by data mining from physicochemical properties.
+#' _Decis. Support Syst._, _47_, 547-553.
+#'
+#' @seealso [loo_subsample()]
+#'
+#' @examples
+#' ## This example predicts wine quality (data from Cortez et al., 2009).
+#' # The code is commented as ## Not run: because brm() takes a few seconds to fit.
+#' # Copy the code to your console to execute it.
+#' # A log_lik_matrix is generated from a fit, then it is used for srs_diff_est().
```
VisruthSK: Hm I don't like how this preamble is worded. Also could you change the `##` to `#`?

Maybe something like 

```r
#' @examples
#' # This example predicts wine quality (data from Cortez et al., 2009).
#' # Examples not run because fitting model is slow.
```

Or maybe don't even explain why the examples are run?

VisruthSK (commented):
Maybe we could add the output of `loo(log_lik_matrix)` commented out to the example too?

avehtari (commented):
I edited the comment in the beginning of the example. Looks good. I would wait for @jgabry's approval when he has time

jgabry:
I’ll try to take a look soon, sorry for the delay.

R/loo_subsample.R
```diff
@@ -1166,12 +1166,86 @@ loo_subsample_estimation_diff_srs <- function(x) {
...
+#'   a proportional difference of \eqn{1/n} compared to the paper.
+#'
+#' @param y_approx (numeric) `n` approximated values.
+#' @param y (numeric) `m<n` subsampled values.
+#' @param y_idx (integerish) The index of `y` in `y_approx`.
+#'
+#' @return A named list containing numeric values:
+#' * `y_hat`: estimated mean of \eqn{y} (Eq 7),
+#' * `v_y_hat`: variance of the mean estimate (Eq 8), and
+#' * `hat_v_y`: estimated variance of \eqn{y} (Eq 9).
```
jgabry, line 1191: @avehtari or @MansMeg please correct me if I'm wrong, but for the documentation of the returned values, wouldn't this be more correct? 

* `y_hat`: estimated total \eqn{\sum_i y_i}, equivalently \eqn{n E[y]}
* `v_y_hat`: estimated sampling variance of `y_hat` due to subsampling
* `hat_v_y`: estimated variance component for the total \eqn{\sum_i y_i} used to compute the standard error `sqrt(hat_v_y)` 

In particular, the current `@return` doc says the estimate `y_hat` is the estimated mean of \eqn{y} but I thought it was the estimate of \eqn{n E[y]}? Or am I mistaken?

R/loo_subsample.R
jgabry, line 1252: Since this function is going to be exported now, should we tighten this check a bit? Currently this permits passing in duplicated values in `y_idx`, and doesn't check max or min value, etc. Maybe something like this? 

```
checkmate::assert_integerish(
  y_idx,
  len = length(y),
  lower = 1,
  upper = length(y_approx),
  unique = TRUE,
  any.missing = FALSE
)
```

R/loo_subsample.R
```diff
@@ -1166,12 +1166,86 @@ loo_subsample_estimation_diff_srs <- function(x) {
...
+#' Magnusson et al. (2020).
+#'
+#' @details In Magnusson et al. (2020) Eq (9) first row, the second `+` should
+#'   be a `-`; Supplementary Material Eq (6) has this correct.
+#'   As `srs_diff_est()` in the `loo` package is used for \eqn{n E[y]}, there is
+#'   a proportional difference of \eqn{1/n} compared to the paper.
+#'
+#' @param y_approx (numeric) `n` approximated values.
+#' @param y (numeric) `m<n` subsampled values.
+#' @param y_idx (integerish) The index of `y` in `y_approx`.
```
jgabry, line 1186:
```suggestion
#' @param y_idx (integerish) The positions in `y_approx` corresponding to the subsampled values in `y`.
```

Maybe this is a bit clearer? Optional change, the current version isn't wrong.

jgabry (changes requested):
Thanks again for working on this! A just made a few suggestions. Sorry for the delay. In addition to the review comments, can you also regenerate the .Rd files? I think there are changes to the doc since those files were last generated. (Although maybe wait to regenerate until any further edits to the doc are made based on my review comments.)

vinniott:
Thank you for the review, @jgabry !
I am unsure: Should I already adopt your suggestions or first wait for input from @avehtari or @MansMeg ?

Also, please excuse my delayed reply; I currently have a few things to take care of.

jgabry:
Thanks! For at least that first suggestion (rewording the doc), let's wait for @avehtari or @MansMeg to confirm. I'll ping them again.

R/loo_subsample.R
jgabry: @MansMeg are you able to comment on this? I think @avehtari is traveling. I'm not entirely sure if my suggested changes are correct or if we should leave the current text. Thanks!

avehtari: There is a mismatch between the paper and the code in whether including n or not, but the code has a comment
> Here the variance is for sum, while in the paper the variance is for mean
> which explains the proportional difference of 1/N

which means you are correct and I like your suggestion

R/loo_subsample.R
avehtari: OK

VisruthSK:
@vinniott Hi! Looks like this PR is close to being good to go. Could you implement the two suggestions Jonah made?

vinniott:
Thanks @VisruthSK for the ping! I hope to get to it timely.
(I know its not a lot to wrap it up, just need to find some headspace between some upcoming deadlines and am looking forward to browse bayesplot afterwards)

PR #352, opened 2026-04-08
Use touchstone

### LINKED ISSUE #348: Performance benchmarks for PRs?
Should we set something up so that we can detect performance regressions introduced by PRs? For example, suppose that one of the PRs to switch to using functions from posterior (e.g. #290, #305) actually caused running loo() to be much slower because the version in posterior was less efficient (just using as an example, I don't have evidence this is true). Currently we have no way of detecting that when reviewing the PR.

VisruthSK:
Use touchstone to evaluate PRs' impact on performance. Starting with a monolithic approach to testing, calling just `loo()`.

Closes #348.


VisruthSK:
Should lower reps to maybe 5, perhapsalso use smaller data? Current ones are fully LLM generated so maybe more careful creation could cut down on execution time while still being informative tests.

@jgabry any thoughts? The main file is touchstone/script.R The code rn is ugly, I'll fix it tomorrow probably, but any thoughts on the actual tests? The run is taking 1hr, is that way too slow?

jgabry:
If we switch to the example that @avehtari suggested, is that a bigger log lik matrix/array than what you have here or smaller?

VisruthSK:
Wine one is a decent bit larger. How long do you figure is too long?

jgabry:
Is there an easy way to tell it to skip the touchdown step? For important PRs it’s fine for it to be slow, even very slow, but for simple PRs that we know won’t affect runtime at all (or trivially) it would be good to be able to skip it. I guess we could always just merge the PR without waiting for it to finish, but sometimes we’re not checking the CI results immediately and then we’d end up just wasting a bunch of computation which I think might prevent other things in the Stan org from running.

avehtari:
Storing that log_lik_matrix and using the stored matrix for loo takes a couple seconds. I don't think that's too slow.

jgabry:
I agree, that’s not too slow. Maybe I’m just confused. I thought that example was bigger than the one Visruth is currently using, and the current run takes 1hr. It runs things many times to compute the benchmarks, but running something that takes a few seconds many times should still take way less than an hour. So then why does the current example take 1hr? Sorry if I’m missing something simple here.

VisruthSK:
I think the LLM code is doing the wrong thing--I'll run the wine thing and store the log lik (probably in the touchstone dir?) and just read from it and run loo. I think the LLMd code is just abysmal and so is too slow. Will try to get this in by Monday.

jgabry:
> I think the LLM code is doing the wrong thing

It might be, I haven't had time to dig into it. Either way, thanks for setting this up. I took a quick look, and I do like that it's also testing the `loo.function` method in addition to the matrix and array methods. But yeah, let's use @avehtari's example.

avehtari:
I tested the current benchmark code by running it manually on my laptop, and it takes less than 10s

VisruthSK:
I looked into the logs and Aki is right--the code is fine. I think the first run was slow because it needed to install R, deps, etc. Did a rerun and it took [5 min ](https://github.com/stan-dev/loo/actions/runs/24173340905/job/70920042345). If the code examples look reasonable could be good to merge. I think the commenting only works when merged.

avehtari:
I think the log_liks the current code generates are bad, there are tens of warnings. Please use the log_lik I suggested, as I know that it's based on real model, has nice properties, and can be used also for srs_diff_est example

VisruthSK:
Swapped to wine data, but didn't write a helper to expose the data, just accessing it through loo:::.wine_log_lik_matrix. The second test is still AI generated. Touchstone won't work since the wine data isn't on master.

jgabry:
On the master branch sysdata.R is <250 KB but now it's almost 40 MB. That's going to be an issue for CRAN I think. Is the new example really that big? If so, then we probably can't include it with the package but we could still put it in the repo and use it here for benchmarking.

VisruthSK:
Maybe I did something wrong but when I saved the loglik to RDS by itself its 39.5 MB. Forgot CRAN had limits on that. If we put the file in the touchstone directory it should be fine right?

Edit: putting in touchstone dir and excluding from build means we can't uses it directly in example, but maybe that's okay?

jgabry:
If we want to use it in examples too then we can create it again in the example by fitting the brms model. Creating a log lik matrix of that size should be ok from CRAN's perspective, just not storing it in the package. The policy is: 

> As a general rule, neither data nor documentation should exceed 5MB 

They do offer exceptions sometimes (RStan is huge), but they are pretty strict with the exceptions so I don't think it's worth asking in a case like this.

jgabry:
And I don't think you did anything wrong necessarily. 40MB is plausible.

VisruthSK:
I'll see how to make the file only available to touchstone then.

VisruthSK:
Right that's all sorted out then. I think if the wine data is changed in a branch, that will be used for that branch's tests too--so just another thing to watch and ensure isn't modified accidentally.

Do we need more tests? Here's what the latest message looked like:

This is how benchmark results would change (along with a 95% confidence interval in relative change) if 4e9c78ef61072798fc69e503cf432c3b1c8b1f47 is merged into master:
* :heavy_check_mark:loo_function: 1.99s -> 2s [-0.14%, +0.98%]
* :heavy_check_mark:loo_matrix: 1.53s -> 1.53s [-0.7%, +0.26%]
Further explanation regarding interpretation and methodology can be found in the [documentation](https://lorenzwalthert.github.io/touchstone/articles/inference.html).

.github/workflows/touchstone-comment.yaml
```diff
@@ -0,0 +1,22 @@
+name: Continuous Benchmarks (Comment)
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
```
jgabry, line 4: Codex seems to think there's a problem with this (I have no idea): 

.github/workflows/touchstone-comment.yaml:4 uses ${{ github.head_ref }} in a workflow triggered by workflow_run. github.head_ref is only defined for pull_request / pull_request_target events. GitHub’s concurrency docs explicitly say to use a fallback when a property is event-specific. As written, the comment workflow can fail before it starts, which means the PR may never post benchmark comments. Suggested fix: use ${{ github.workflow }}-${{ github.head_ref || github.run_id }}.

.github/workflows/touchstone-comment.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  upload:
+    runs-on: ubuntu-latest
+    if: >
+      ${{ github.event.workflow_run.event == 'pull_request' }}
+    steps:
+      - uses: lorenzwalthert/touchstone/actions/comment@main
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jgabry: Do we need to set up a token for this?

touchstone/script.R
```diff
@@ -0,0 +1,57 @@
...
+  expr_before_benchmark = {
+    suppressPackageStartupMessages(library(loo))
+    # benchmark_run() evaluates in a callr subprocess, so load pinned assets here.
+    wine_log_lik_matrix <- readRDS(touchstone::path_pinned_asset(
+      "touchstone/wine.rds"
+    ))
+    matrix_r_eff <- rep(1, ncol(wine_log_lik_matrix))
+  },
+  loo_matrix = {
+    suppressWarnings(
```
jgabry, line 21: Do we want to suppressWarnings, or can we allow warnings and see if the new code changes the warnings or the number of warnings? Or is that outside the scope of the touchstone functionality?

touchstone/script.R
```diff
@@ -0,0 +1,57 @@
...
+    suppressPackageStartupMessages(library(loo))
+    wine_log_lik_matrix <- readRDS(touchstone::path_pinned_asset(
+      "touchstone/wine.rds"
+    ))
+    function_r_eff <- rep(1, ncol(wine_log_lik_matrix))
+    wine_data <- data.frame(obs = seq_len(ncol(wine_log_lik_matrix)))
+    wine_llfun <- function(data_i, draws) draws[, data_i$obs, drop = FALSE]
+  },
+  loo_function = {
+    suppressWarnings(
```
jgabry, line 43: Same question about warnings as above

.github/workflows/touchstone-comment.yaml
jgabry: I'm not sure if this is actually an issue. Is this just the standard yaml file used for this or did you customize it at all?

VisruthSK: Yeah these are (mostly) the standard YAMLs that `touchstone::use_touchstone()` generate. They seem out of date though--the actual action is pinned to v1 even though it should be to main AFAIK, so its possible this is a real issue. What Codex says vaguely makes sense, but I think we could just remove workflow_run instead of acommodating it? I think workflow_run is [supposedly dangerous anyway](https://astral.sh/blog/open-source-security-at-astral#cicd-security)

Edit: okay so we can't remove workflow_run. I don't think its an issue. touchstone works in stanflow already with the same setup.

.github/workflows/touchstone-comment.yaml
VisruthSK: As in, can we get away with not using it? Not sure, same as above this is the default. Can remove and try to see if works?

touchstone/script.R
VisruthSK: I think that's out of scope--I think touchstone only cares about performance. Not exactly sure how benchmark_run() handles warnings but I don't think silencing warnings is too bad.

jgabry: Ok yeah. It could be nice to know if the change in a PR leads to a lot more warnings, but if that's outside the scope of touchstone then that's fine

.github/workflows/touchstone-comment.yaml
jgabry: I would think we could just explicitly set a few permissions, e.g. 

```
  permissions:
    contents: read
    pull-requests: write
    issues: write
```

but it's not a big deal. I think it's fine to use the token. Also maybe I'm misunderstanding how this work.

VisruthSK: Ahh right, no I think you're right. The default GHA is a touch stale and I think we should be doing things like this to make it slightly more secure.

touchstone/script.R
VisruthSK: Snapshots would cover this to a degree too

.github/workflows/touchstone-comment.yaml
VisruthSK: Modified this, I hope the actual commenting works, but can only check that after merging :)

jgabry (approved):
Let's go ahead and merge this so we can try it out and then we can make modifications as needed. @VisruthSK should we make a test PR after merging this to see if the commenting works? Or just try it with one of the open PRs?

VisruthSK:
Might need to make a test pr, but updating one of of the old PRs might be okay.

VisruthSK:
https://github.com/stan-dev/loo/pull/305#issuecomment-4247447968

Cool looks like its working!

jgabry:
Nice!

PR #356, opened 2026-04-16
Increase touchstone iterations

VisruthSK:


VisruthSK:
I'm confused as to why the system deps are now taking forever to install. I mean before, on a rerun, n=10 (so 20 benchmarks, 10 on main, 10 on branch) took only 5 minutes. Now its already been almost 15 min and the GHA is still installing system dependencies.

Edit: took about 16 minutes before the actual touchstone runs started.

Issue #343, opened 2026-03-28
Feature Request: public subset() method for psis objects


utkarshpawade:
There is currently no public API in the `loo` package to subset a `psis` object by observation index. Downstream packages that need this functionality are forced to manually manipulate the internal structure of `psis` objects.

For example, bayesplot has had this workaround sitting with a FIXME comment : https://github.com/stan-dev/bayesplot/blob/master/R/ppc-loo.R#L722-L738

Would it be possible to export a `subset.psis()` S3 method or a `subset_psis()` helper? That way downstream code can drop the fragile workaround and rely on a stable API.

Issue #179, opened 2021-06-12
Calibration/Refinement Decomposition of LOO Scores


ParadaCarleton:
I think a breakdown of LOO scores into calibration/refinement would be interesting and useful to users. Mentioning @avehtari because Jonah mentioned you'd proposed adding this decomposition and said you might have some related work.

ParadaCarleton:
It has now been 2 years and I have no idea what I was talking about then :smile:

But I think [this](https://link.springer.com/content/pdf/10.1007/978-3-319-23528-8_5.pdf) was the relevant paper that inspired this suggestion, though I'm not 100% sure.

PR #363, opened 2026-06-02
Adding `pred_measure` feature

### LINKED ISSUE #135: user defined loss/utility functions
Section 4.3 of https://avehtari.github.io/modelselection/diabetes.html demonstrates how to use loo package with other loss/utility function than log score. LOO-R^2 is another example. Both of these are just for point predictions, which is the easiest case.

Make it easier to provide user defined loss/utility function for point predictions. Arguments would be
 - point estimate type given for E_loo (e.g. "mean")
 - user defined function with at least two arguments 1) point predictions and 2) data y

We can later consider expanding to loo/utility functions for predictive distributions.

### LINKED ISSUE #201: Allow use of other criteria

### LINKED ISSUE #213: loo_predictive_metric and loo_crps could accept psis objects to reduce computation
If the user already has a psis object then `loo_predictive_metric` and `loo_[s]crps` could use that object instead of rerunning psis internally (similar to what `E_loo` already does, except in this case it would be optional). If no psis object is provided then they could compute it as they do now. @LeeviLindgren @avehtari what do you think? (This doesn't need to be done right now, it could be for the next release after this one.)

### LINKED ISSUE #220: loo_compare for crps and loo_crps
I just discovered `crps`! This is great, I was actually asking about how to compare probabilistic vs non-probabilistic predictions https://discourse.mc-stan.org/t/comparison-of-machine-learning-models-and-bayesian-models/30833 and this is exactly what I needed.

I just noticed that `loo_compare` and `print` are not implemented. If it's ok, I volunteer to implement it.  I just want to know how would you call the predictions, "elpd" doesn't make much sense. Maybe "ps"? (predictive score)?

### LINKED ISSUE #223: (loo_)(s)crps could ask for only one argument with predictions
I think it would make sense to ask the user  for only one argument with predictions with at least two columns and split the predictions array. (Also in  #221).

In my tests, if you just divide the predictions in two, the estimates are virtually the same:
```r
(ps2 <- crps(ypred1[1:2000,], ypred1[2001:4000,], y = fit$y))
```

@avehtari said:
> I think it would be clearer to have this discussed in a separate issue (and tag @LeeviLindgren). I don't expect there is much difference in the easy case, but if the variability of the importance weights is big then a) the importance weights are different for the two halves (while in the current implementation they are the same), b) the number of draws for the importance weighting is halved, which increases Monte Carlo error. It is possible that in practice a + b are not a problem, but this should be investigated more thoroughly.

### LINKED ISSUE #281: New functions for better support for different scores and metrics
This is related to issues #223, #220, and #201

We have discussed rewriting CRPS and SCRPS functions to use better computation (probability weighted moment form). As new functions are needed, this has lead to thinking overall improvement of metric and score functions. Some thoughts

- Keep the current loo for elpd, as it does not have arguments for the predictions and observations
- a) Deprecate loo_predictive_metric and create new loo_predictive_measure, or
  b) create new loo_predictive_score
   - (a) might be better if we change the output
   - Noa thinks we could drop loo_
   - I think we could drop predictive_ but should not drop loo_
- For all measures, return a loo object with pointwise and estimates
   - With pointwise information, we can do model comparisons
   - Should we extend the current loo object or make a new object type?
   - One challenge might be the sub sampling loo, which increase the amount of
     of work when implementing other measures
- Extend loo_compare to work with other measures, as we know how to compute diff_se for all current
  metrics and scores. Currently, loo_predictive_metric returns only estimate and se for one model
  predictions
    - Or do we need a new function
- Include MAE, RMSE, MSE, R2, ACC, balanced ACC, and Brie score to metrics
- Include RPS, SRPS, CRPS, SCRPS, and log score to scores
- When computing S?C?RPS or current metrics, maybe store function specific diagnostics?
- All measures need psis object or log weights, and thus
  should we always compute p_loo, too? Or should we compute measure specific value describing
  amount of fitting?
- I have a minimal function that computes loo versions of RPS, SRPS, CRPS, SCRPS and includes documentation for the equations with references (but no argument checking, and computes it only for one scalar y, etc)

tagging @n-kall and @jgabry

florence-bockting:
Fixes #281

## Summary
Adds a unified **predictive performance measures** API on top of `loo-v3.0.0`. Density scores, scoring rules, and point-prediction metrics can be computed under in-sample, PSIS-LOO, k-fold, and holdout evaluation via `*_pred_measure()`.

**Design log:** [`developer-notes.md`](https://github.com/stan-dev/loo/blob/pred_measure/notes/developer-notes.md) (edit: [`developer-notes.Rmd`](https://github.com/stan-dev/loo/blob/pred_measure/notes/developer-notes.Rmd))  

## Issues addressed
This PR consolidates work that was tracked across several related issues:

- [x] Fixes #223: scoring rules use a single `ypred` draw matrix (`measure_rps()`, `measure_srps()`)
- [x] Fixes #213: `loo_pred_measure()` accepts a pre-computed `psis_object` (or extracts one from a `loo` result with `save_psis = TRUE`)
- [x] Fixes #201 / #135: custom measures via user-supplied functions; `measure_r2()` as a built-in
- [x] Fixes #220 (partial): CRPS/RPS and other measures under in-sample, LOO, k-fold, and holdout via `*_pred_measure()` — **`loo_compare` integration still outstanding** (see [Known limitations](#known-limitations))

## How to navigate this PR
This is a large PR. These documents explain design, scope, and current status:

- [`developer-notes.md`](https://github.com/stan-dev/loo/blob/add-pred-measure/notes/developer-notes.md):
  central reference: status, open tasks, design decisions, numerical comparisons
- [`overview-measures.Rmd`](https://github.com/stan-dev/loo/blob/add-pred-measure/vignettes/articles-online-only/overview-measures.Rmd): supported scores and metrics, with mathematical definitions
- [`pred-measure-workflow.Rmd`](https://github.com/stan-dev/loo/blob/add-pred-measure/vignettes/articles-online-only/pred-measure-workflow.Rmd): conceptual introduction and end-to-end workflow examples
- [`migration-guide.Rmd`](https://github.com/stan-dev/loo/blob/add-pred-measure/vignettes/migration-guide.Rmd): mapping from deprecated functions to the new API

## What's new
- **Entry points:** `insample_pred_measure()`, `loo_pred_measure()`, `kfold_pred_measure()`, `test_pred_measure()`, `pred_measure()`
- **Built-in measures:** `measure_*()` + `supported_measures_list()` (incl. `r2`, classification metrics, Brier, MAE/MSE/RMSE, ELPD/IC)
- **Unified scoring rule:** `measure_rps()` / `measure_srps()` — single `ypred` matrix, PWM/ECDF estimator; covers continuous and ordered categorical outcomes (CRPS/RPS and scaled variants in the new API)
- **Custom measures:** pass a function (or named list of functions) to `measure`
- **PSIS reuse:** supply `psis_object` or reuse weights from a prior `loo_pred_measure()` result via `predperf`
- **S3 print methods** for `pred_measure` objects (incl. Pareto-k-hat for LOO)
- **Docs:** `migration-guide` vignette + website-only `overview-measures` and `pred-measure-workflow` articles
- **Tests & CI:** test suite, pre-fitted `.Rds` fixtures, pkgdown vignette-data generation step

## Deprecations (implementations retained)
Existing functions are deprecated but **still work with their current APIs**:

| Deprecated | New workflow |
|------------|--------------|
| `elpd()` | `measure_elpd()` / `*_pred_measure()` |
| `crps()`, `scrps()` (`x`, `x2`) | `measure_rps()`, `measure_srps()` (`ypred`) |
| `loo_crps()`, `loo_scrps()` | `loo_pred_measure(..., measure = "rps" \| "srps")` |
| `loo_predictive_metric()` | `loo_pred_measure()` |

In `*_pred_measure()`, `elpd` is computed automatically when `ylp` is supplied; request `measure = "ic"` separately for the information criterion.

The new RPS functions use a different estimator (one draw matrix, PWM) than deprecated CRPS (two draw matrices, permutation). Results are **highly correlated but not identical** — see the comparison section and figures in [`developer-notes.Rmd`](../notes/developer-notes.Rmd). Comparison tests are in `tests/testthat/test_crps.R`.

## Dependencies
Vignettes, test-data generation, and pkgdown CI install **brms from GitHub `master`** (`paul-buerkner/brms`), because k-fold support for categorical / multinomial models ([brms#1890](https://github.com/paul-buerkner/brms/pull/1890), fixing [brms#1889](https://github.com/paul-buerkner/brms/issues/1889)) is not yet on CRAN. Once a CRAN release includes that fix, the GitHub pin can be dropped.

## Known limitations
- `group_ids` grouping not yet implemented
- `loo_compare` integration still outstanding
- Additional documentation still in progress (formula derivations article, glossary extension)
- brms GitHub pin remains until a CRAN release includes the categorical k-fold fix

See [`developer-notes.Rmd`](../notes/developer-notes.Rmd) for open design questions and
remaining tasks.

## Acknowledgements
This work builds on an initial prototype of the `pred_measure` feature implemented by @VisruthSK (see [VisruthSK/loo-sandbox](https://github.com/VisruthSK/loo-sandbox)) under guidance from @jgabry.

## Test plan
- [x] `devtools::test()` and `devtools::check()` pass
  - `devtools::test()`: 1 failure — unrelated `test_psislw.R` snapshot drift (`Warning in psislw()` → `Warning:`)
  - `R CMD check`: 1 error — `kfold_pred_measure()` roxygen example uses `measure = "rmse"` without `mupred`
- [x] Deprecated functions warn (`test_deprecated_measures.R`)
- [x] Deprecated functions return same results *(snapshot tests in `test_crps.R`; not re-diffed against `loo-v3.0.0` branch)*
- [x] CRPS/RPS comparison tests pass (`test_crps.R`, 32 tests)
- [x] New `*_pred_measure()` examples run (migration guide + workflow vignette; rendered locally with `NOT_CRAN=true`)
- [ ] pkgdown site builds with GitHub `brms` *(workflow + migration-guide articles render; full `pkgdown::build_site()` not verified locally — existing `docs/` blocks build)*


florence-bockting:
Linking here to Issue #179 as it seems to be about loo scores. We should see whether it is still relevant and can be implemented within loo refactoring.

- [ ] Add issue #179 to PR description text if we can address it.

florence-bockting:
There is an open issue requesting a subset() method for psis objects #343 and an (unrelated) open PR #110. Might be worth considering whether we can address this as well in this refactoring PR. If yes, we should 

- [x] close PR #110 with link to this PR and also 
- ~[ ] mark issue #343 as fixed in the description text of this PR.~

Moved this to PR #379

PR #374, opened 2026-06-10
Mention new loo_compare columns in vignette

jgabry:
Briefly mention new `loo_compare` output in the main vignette. For now I just point to the glossary and Aki and Florence's case study.


florence-bockting (approved):
Looks good, thank you @jgabry.

